PR #2, opened 2019-01-28
Release via gradle-release with a secret GitHub token

dagguh:
I'm trying to mirror our current workflow from Bitbucket, where we have a manual step/button to trigger the release from CI. Currently on Travis it might be too easy to accidentally trigger it, e.g. here: https://travis-ci.com/atlassian/jira-actions/builds/98801634#L744

TBH we should be more automated, e.g. release if there's a Git tag present or something like that, but I'll leave that for later


.travis.yml
```diff
@@ -1,10 +1,21 @@
 language: java
 
 stages:
+  - test
   - name: release
-    if: branch == master AND fork == false AND type == api
+    if: env(RELEASE) == true AND fork == false
```
dagguh: Hmm, we could actually get away with this:
```suggestion
    if: env(RELEASE) == true
```
The permissions are inherently checked by Artifactory and GitHub, we don't actually need to filter out forks here.
OTOH it might be worth leaving it as is, just to show the intent of the workflow.

dagguh:
I had to add 6c004ba678b076509bc52e7700e1b4aa2a88618a, because we uncovered a hidden gradle-release dependency on the Gradle API

PR #3, opened 2019-02-19
Updated the EditPage and CommentInsert to support rich text editor

kema2000:
1. indicates if rich text is enabled or not based on CSS
2. If rich text is enabled, navigate to the sub iFrame and update the field
3. otherwise update the text-area directly

### DIFF
--- a/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/CommentForm.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/CommentForm.kt
@@ -34,7 +34,8 @@ class CommentForm(
     fun enterCommentText(
         comment: String
     ): CommentForm {
-        driver.findElement(By.id("comment")).sendKeys(comment)
+        RichTextEditorTextArea(driver, driver.findElement(By.id("comment")))
+            .overwriteIfPresent(comment)
         return this
     }
 }
--- a/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/EditIssuePage.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/EditIssuePage.kt
@@ -1,17 +1,16 @@
 package com.atlassian.performance.tools.jiraactions.api.page
 
 import com.atlassian.performance.tools.jiraactions.page.form.IssueForm
-import org.openqa.selenium.By
-import org.openqa.selenium.WebDriver
+import org.openqa.selenium.*
 import org.openqa.selenium.support.ui.ExpectedConditions.*
 import java.time.Duration
+
 
 class EditIssuePage(
     private val driver: WebDriver
 ) {
     private val form = IssueForm(By.id("issue-edit"), driver)
     private val summaryLocator = By.id("summary")
-    private val descriptionLocator = By.id("description")
     private val updateButtonLocator = By.id("issue-edit-submit")
 
     fun waitForEditIssueForm(): EditIssuePage {
@@ -31,7 +30,11 @@ class EditIssuePage(
 
     fun fillForm(): EditIssuePage {
         summaryLocator.clearAndTypeIfPresent("summary")
-        descriptionLocator.clearAndTypeIfPresent("description")
+        val descriptionFieldId = "description"
+        if(driver.isElementPresent(By.id(descriptionFieldId))){
+            RichTextEditorTextArea(driver, driver.findElement(By.id(descriptionFieldId)))
+                .overwriteIfPresent("description")
+        }
         form.fillRequiredFields()
         return this
     }
@@ -49,4 +52,5 @@ class EditIssuePage(
             webElement.sendKeys(text)
         }
     }
+
 }
--- a/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/RichTextEditorTextArea.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/RichTextEditorTextArea.kt
@@ -1,0 +1,74 @@
+package com.atlassian.performance.tools.jiraactions.api.page
+
+import org.apache.logging.log4j.LogManager
+import org.apache.logging.log4j.Logger
+import org.openqa.selenium.By
+import org.openqa.selenium.WebDriver
+import org.openqa.selenium.WebElement
+import org.openqa.selenium.support.ui.ExpectedConditions.*
+import java.time.Duration
+
+
+
+/**
+ *
+ * represents the [Jira Rich Text Editor](https://confluence.atlassian.com/adminjiraserver/rich-text-editing-938847886.html)
+ * falls back to the plain text editor if RTE is disabled
+ *
+ */
+class RichTextEditorTextArea(
+    private val driver: WebDriver,
+    private val textArea: WebElement
+) {
+
+    private val logger: Logger = LogManager.getLogger(this::class.java)
+
+    fun overwriteIfPresent(text: String) {
+        val attrClass = textArea.getAttribute("class")
+        if(attrClass != null && attrClass.contains(other = "richeditor-cover", ignoreCase = true)){
+            overwriteRich(text)
+        }
+        else{
+            overwritePlain(text)
+        }
+    }
+
+    private fun overwritePlain(text: String) {
+        driver.wait(timeout = Duration.ofSeconds(2), condition = elementToBeClickable(textArea))
+        textArea.clear()
+        textArea.sendKeys(text)
+    }
+
+    private fun overwriteRich(text: String) {
+        /**
+         * the tinymce Iframe element structure:
+         * div
+         *   -- textarea
+         *   -- div
+         *     -- div
+         *       -- div
+         *         -- div
+         *           --iframe
+         */
+        val iframeXpath = "//div[textarea[@id='${textArea.getAttribute("id")}']]//iframe"
+
+        try {
+            driver.wait(
+                timeout = Duration.ofSeconds(10),
+                condition = frameToBeAvailableAndSwitchToIt(By.xpath(iframeXpath))
+            )
+            //tinymce uses body element with id 'tinymce' to hold the text, find the body element
+            val tinyMce = driver.findElement(By.id("tinymce"))
+            //data-projectkey={project}
+            driver.wait(
+                timeout = Duration.ofSeconds(10),
+                condition = attributeToBeNotEmpty(tinyMce, "data-projectkey")
+            )
+            tinyMce.clear()
+            tinyMce.sendKeys(text)
+        } finally {
+            driver.switchTo().parentFrame()
+        }
+    }
+
+}
--- a/src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/RichTextEditorIT.kt
+++ b/src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/RichTextEditorIT.kt
@@ -1,0 +1,163 @@
+package com.atlassian.performance.tools.jiraactions.api.scenario
+
+import com.atlassian.performance.tools.dockerinfrastructure.api.browser.DockerisedChrome
+import com.atlassian.performance.tools.dockerinfrastructure.api.jira.JiraCoreFormula
+import com.atlassian.performance.tools.jiraactions.api.ActionMetric
+import com.atlassian.performance.tools.jiraactions.api.ActionResult
+import com.atlassian.performance.tools.jiraactions.api.SeededRandom
+import com.atlassian.performance.tools.jiraactions.api.WebJira
+import com.atlassian.performance.tools.jiraactions.api.action.*
+import com.atlassian.performance.tools.jiraactions.api.measure.ActionMeter
+import com.atlassian.performance.tools.jiraactions.api.measure.output.CollectionActionMetricOutput
+import com.atlassian.performance.tools.jiraactions.api.memories.User
+import com.atlassian.performance.tools.jiraactions.api.memories.UserMemory
+import com.atlassian.performance.tools.jiraactions.api.memories.adaptive.AdaptiveIssueKeyMemory
+import com.atlassian.performance.tools.jiraactions.api.memories.adaptive.AdaptiveIssueMemory
+import com.atlassian.performance.tools.jiraactions.api.memories.adaptive.AdaptiveJqlMemory
+import com.atlassian.performance.tools.jiraactions.api.memories.adaptive.AdaptiveProjectMemory
+import com.atlassian.performance.tools.jiraactions.api.w3c.DisabledW3cPerformanceTimeline
+import org.assertj.core.api.Assertions
+import org.junit.Test
+import java.time.Clock
+import java.util.*
+
+/**
+ * Test Editing issue without SetupAction (i.e Rich Text editor enabled)
+ * Covers:
+ *      EditIssue
+ *      AddComment
+ */
+class RichTextEditorIT {
+
+    @Test
+    fun shouldRunScenarioWithoutErrors() {
+
+        /**
+         * Original intention was to test multiple JIRA version in one test case.
+         *
+         * This is not working now due to:
+         * 1. the static network alias is used
+         * 2. the jira.close() does not release the resource properly yet
+         *
+         * So we will keep it for one version per run now.
+         */
+        val versions = listOf(
+            "8.0.0"
+//            "7.13.0"
+//            "7.3.0"
+//            "7.2.0"
+        )
+        var basePort = 8080;
+
+        for (version in versions) {
+            runTestOnOneInstance(version, basePort++)
+        }
+    }
+
+    private fun runTestOnOneInstance(version: String, port : Int) {
+        println("Testing JIRA with port : $port, version : $version")
+        val scenario = JiraEditScenario()
+        val metrics = mutableListOf<ActionMetric>()
+        val actionMeter = ActionMeter(
+            virtualUser = UUID.randomUUID(),
+            output = CollectionActionMetricOutput(metrics),
+            clock = Clock.systemUTC(),
+            w3cPerformanceTimeline = DisabledW3cPerformanceTimeline()
+        )
+        val user = User("admin", "admin")
+        val userMemory = object : UserMemory {
+            override fun recall(): User {
+                return user
+            }
+
+            override fun remember(memories: Collection<User>) {
+                throw Exception("not implemented")
+            }
+        }
+        JiraCoreFormula.Builder()
+            .version(version)
+            .port(port)
+            .build()
+            .provision()
+            .use { jira ->
+                DockerisedChrome().start().use { browser ->
+                    val webJira = WebJira(
+                        browser.driver,
+                        jira.getUri(),
+                        user.password
+                    )
+                    val logInAction = scenario.getLogInAction(
+                        webJira,
+                        actionMeter,
+                        userMemory
+                    )
+                    val actions = scenario.getActions(
+                        webJira,
+                        SeededRandom(123),
+                        actionMeter
+                    )
+
+                    logInAction.run()
+                    actions.forEach { action ->
+                        action.run()
+                    }
+                }
+                jira.close()
+            }
+
+        val results = metrics.map { metric ->
+            metric.result
+        }
+        Assertions.assertThat(results).containsOnly(ActionResult.OK)
+    }
+}
+
+class JiraEditScenario constructor() : Scenario {
+
+    override fun getActions(jira: WebJira, seededRandom: SeededRandom, meter: ActionMeter): List<Action> {
+        val projectMemory = AdaptiveProjectMemory(random = seededRandom)
+        val jqlMemory = AdaptiveJqlMemory(seededRandom)
+        val issueKeyMemory = AdaptiveIssueKeyMemory(random = seededRandom)
+        val issueMemory = AdaptiveIssueMemory(issueKeyMemory, seededRandom)
+
+        val createIssue = CreateIssueAction(
+            jira = jira,
+            meter = meter,
+            seededRandom = seededRandom,
+            projectMemory = projectMemory
+        )
+        val searchWithJql = SearchJqlAction(
+            jira = jira,
+            meter = meter,
+            jqlMemory = jqlMemory,
+            issueKeyMemory = issueKeyMemory
+        )
+        val viewIssue = ViewIssueAction(
+            jira = jira,
+            meter = meter,
+            issueKeyMemory = issueKeyMemory,
+            issueMemory = issueMemory,
+            jqlMemory = jqlMemory
+        )
+        val editIssue = EditIssueAction(
+            jira = jira,
+            meter = meter,
+            issueMemory = issueMemory
+        )
+        val addComment = AddCommentAction(
+            jira = jira,
+            meter = meter,
+            issueMemory = issueMemory
+        )
+
+        return mutableListOf(
+            createIssue,
+            searchWithJql,
+            viewIssue,
+            editIssue,
+            addComment
+        )
+    }
+
+
+}
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ stages:
 jobs:
   include:
     - stage: test
-      script: travis_wait 14 ./gradlew build
+      script: travis_wait 21 ./gradlew build
     - stage: release
       script: ./scripts/travis-release.sh
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,11 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://bitbucket.org/atlassian/jira-actions/branches/compare/master%0Drelease-3.3.0
 
+### Fixed - 2019-02-27
+- Support RTE with a fallback to plain-text editor. Resolve [JPERF-184]
+
+[JPERF-184]: https://ecosystem.atlassian.net/browse/JPERF-184
+
 ## [3.3.0] - 2019-01-29
 [3.3.0]: https://bitbucket.org/atlassian/jira-actions/branches/compare/release-3.3.0%0Drelease-3.2.0
 


src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/JiraRteEditPageTest.kt
```diff
@@ -0,0 +1,174 @@
...
+import java.time.Clock
+import java.util.*
+
+/**
+ * Test Editing issue without SetupAction (i.e Rich Text editor enabled)
+ * Covers:
+ *      EditIssue
+ *      AddComment
+ */
+class JiraRteEditPageTest {
```
dagguh: Huge props for adding a test!

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/JiraRteEditPageTest.kt
```diff
@@ -0,0 +1,174 @@
...
+
+    @Test
```
dagguh: WRT the [failed build](https://travis-ci.com/atlassian/jira-actions/builds/101537582), it didn't fit in the current 14 minute timeout. I'm running the tests locally to see how long does it take and therefore what should the timeout be.

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/JiraRteEditPageTest.kt
```diff
@@ -0,0 +1,174 @@
...
+            meter = meter,
+            issueMemory = issueMemory
+        )
+        val addComment = AddCommentAction(
+            jira = jira,
+            meter = meter,
+            issueMemory = issueMemory
+        )
+
+        val scenario: MutableList<Action> = mutableListOf(
```
dagguh: I like how you narrowed down the action list from `JiraCoreScenario` in order to leave only:
* the RTE actions themselves (`createIssue`, `editIssue`, `addComment`)
* the exploratory actions for `issueMemory` and `issueKeyMemory` (`searchWithJql`, `viewIssue`)

However, it seems we don't need `viewDashboard`.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/EditIssuePage.kt
```diff
@@ -31,7 +34,8 @@ class EditIssuePage(
 
     fun fillForm(): EditIssuePage {
         summaryLocator.clearAndTypeIfPresent("summary")
-        descriptionLocator.clearAndTypeIfPresent("description")
+        //By.id("description").clearAndTypeIfPresent("Test Desc")
```
dagguh: We should keep the codebase clean of dead/commented out code

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/EditIssuePage.kt
```diff
@@ -49,4 +53,52 @@ class EditIssuePage(
             webElement.sendKeys(text)
         }
     }
+
+}
+
+object PageRteUtil{
```
dagguh: So this is a Kotlin `object`, which is basically a singleton, which is basically a global.
Instead of using a global container for functions, we can keep things local via actual object instances. We can follow the page object pattern, just like we do with `IssueCreateDialog`, `TextArea` etc. The Jira editor is just another page component.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/EditIssuePage.kt
```diff
@@ -49,4 +53,52 @@ class EditIssuePage(
...
+            }
+            else{
+                //try plain text editing
+                logger.debug("Try use Raw Text Editer...")
+                clearAndTypeIfPresent(driver, id, text)
+            }
+        }
+    }
+
+    private fun logElement(bodyElement: WebElement, driver: WebDriver) {
```
dagguh: We should clean up the temporary debugging constructs

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/EditIssuePage.kt
```diff
@@ -49,4 +53,52 @@ class EditIssuePage(
...
+            val webElement = driver.findElement(By.id(id))
+            driver.wait(timeout = Duration.ofSeconds(2), condition = elementToBeClickable(webElement))
+            webElement.clear()
+            webElement.sendKeys(text)
+        }
+    }
+
+    fun clearAndTypeIfPresentRte(driver: WebDriver,id : String, text: String) {
+        //check richeditor-cover exists in class attribute
+        if (driver.isElementPresent(By.id(id))) {
```
dagguh: This `if` doesn't have an `else`. So it seems that if we pass a wrong `id` or the page is broken, we'll silently ignore that.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/EditIssuePage.kt
```diff
@@ -49,4 +53,52 @@ class EditIssuePage(
...
         }
     }
+
+}
+
+object PageRteUtil{
+
+    private val logger: Logger = LogManager.getLogger(this::class.java)
+
+    fun clearAndTypeIfPresent(driver: WebDriver, id :String, text: String) {
```
dagguh: This one can be private. We want to only expose the RTE-compatible behavior.

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/JiraRteEditPageTest.kt
```diff
@@ -0,0 +1,174 @@
...
+            metric.result
+        }
+        Assertions.assertThat(results).containsOnly(ActionResult.OK)
+    }
+}
+
+class JiraEditScenario constructor() : Scenario {
+    private lateinit var issueKeyMemory: IssueKeyMemory
+
+    private constructor(issueKeyMemory: IssueKeyMemory?) : this() {
```
dagguh: We can avoid the entire `issueKeyMemory` injection. It's present in `JiraCoreScenario` for API consumers, which want to share memories across scenarios. We don't need that feature in this test.
Let's replace it with a simple local variable.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/EditIssuePage.kt
kema2000: sure, removed

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/EditIssuePage.kt
kema2000: yes, agree

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/EditIssuePage.kt
kema2000: yes

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/EditIssuePage.kt
kema2000: That seems to be a common practice for all other Page Api.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/EditIssuePage.kt
kema2000: yep

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/JiraRteEditPageTest.kt
kema2000: okay, got it

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/RichTextEditorTextArea.kt
```diff
@@ -0,0 +1,59 @@
...
+                driver.wait(timeout = Duration.ofSeconds(10), condition = ExpectedConditions.attributeToBeNotEmpty(bodyElement, "data-projectkey"))
+                bodyElement.clear()
+                bodyElement.sendKeys(text)
+
+                driver.switchTo().parentFrame()
+            }
+            else{
+                //try plain text editing
+                logger.debug("Try use Raw Text Editer...")
+                clearAndTypeIfPresent(id, text)
```
dagguh: This `if`/`else` looks a bit imbalanced.
The `if` has 7 logical lines and works on low-level WebDriver logic.
The `else` has 1 logical line (+1 debugging line) and operates on a high-level page object logic.

I'd pack up the `if` contents into a `private fun` to see basically say:
```kotlin
if (isRte(element)) {
    clearAndTypeRichTextIfPresent(element, text)
} else {
    clearAndTypePlainTextIfPresent(element, text)
}
```

Or if we apply the other suggestions:
```kotlin
if (isRich(element)) {
    overwriteRich(element, text)
} else {
    overwritePlain(element, text)
}
```

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/RichTextEditorTextArea.kt
```diff
@@ -0,0 +1,59 @@
...
+        if (driver.isElementPresent(By.id(id))) {
+            val webElement = driver.findElement(By.id(id))
+            val attrClass = webElement.getAttribute("class")
+            if(attrClass != null && attrClass.contains(other = "richeditor-cover", ignoreCase = true)){
+                logger.debug("richeditor detected, will try Rich Text Editer...")
+                val iframeXpath = "//div[textarea[@id='$id']]//iframe"
+                driver.wait(timeout = Duration.ofSeconds(10), condition = frameToBeAvailableAndSwitchToIt(By.xpath(iframeXpath)))
+
+                val bodyElement = driver.findElement(By.id("tinymce"))
+                //data-projectkey={project}
```
dagguh: What does this mean?

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/RichTextEditorTextArea.kt
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    fun clearAndTypeIfPresentRte(id : String, text: String) {
+        //check richeditor-cover exists in class attribute
+        if (driver.isElementPresent(By.id(id))) {
+            val webElement = driver.findElement(By.id(id))
+            val attrClass = webElement.getAttribute("class")
+            if(attrClass != null && attrClass.contains(other = "richeditor-cover", ignoreCase = true)){
+                logger.debug("richeditor detected, will try Rich Text Editer...")
+                val iframeXpath = "//div[textarea[@id='$id']]//iframe"
```
dagguh: Is `//div[textarea[@id='$id']]` the same as `webElement`?
Would `webElement.findElement(By.tagName("iframe")` work?

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/RichTextEditorTextArea.kt
```diff
@@ -0,0 +1,59 @@
...
+        }
+    }
+
+    fun clearAndTypeIfPresentRte(id : String, text: String) {
+        //check richeditor-cover exists in class attribute
+        if (driver.isElementPresent(By.id(id))) {
+            val webElement = driver.findElement(By.id(id))
+            val attrClass = webElement.getAttribute("class")
+            if(attrClass != null && attrClass.contains(other = "richeditor-cover", ignoreCase = true)){
+                logger.debug("richeditor detected, will try Rich Text Editer...")
```
dagguh: This debug logging basically marks the entry into the `if`. Let's say there's a bug somewhere within the `if` and an exception is thrown. Its stacktrace will already prove that it failed within the `if`. Seems unnecessary and spammy, because it (or its `else` counterpart` will be logged for every action using Jira text field.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/RichTextEditorTextArea.kt
```diff
@@ -0,0 +1,59 @@
...
+            val webElement = driver.findElement(By.id(id))
+            driver.wait(timeout = Duration.ofSeconds(2), condition = elementToBeClickable(webElement))
+            webElement.clear()
+            webElement.sendKeys(text)
+        }
+    }
+
+    fun clearAndTypeIfPresentRte(id : String, text: String) {
+        //check richeditor-cover exists in class attribute
+        if (driver.isElementPresent(By.id(id))) {
```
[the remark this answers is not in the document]
dagguh: I found that `EditIssuePage` does that:
```kotlin
summaryLocator.clearAndTypeIfPresent("summary")
descriptionLocator.clearAndTypeIfPresent("description")
form.fillRequiredFields()
```
However, it makes sense there, because:
* summary and description might actually be absent, we've got a customer with such setup
* we at least the required fields, ensuring the edit dialog is submittable

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/RichTextEditorTextArea.kt
```diff
@@ -0,0 +1,59 @@
...
+ * Helper class for Rich Text Editor enabled Textarea
+ *
+ */
+class RichTextEditorTextArea(
+    private val driver: WebDriver
+) {
+
+    private val logger: Logger = LogManager.getLogger(this::class.java)
+
+    private fun clearAndTypeIfPresent(id :String, text: String) {
```
dagguh: So this named consistently with `EditIssuePage` internals, which by itself is good.
However, I think we set up a bad example there. According to [SRP](https://en.wikipedia.org/wiki/Single_responsibility_principle), methods should be responsible for one thing only. Any time you see `and`, you should become suspicious - it has multiple responsibilities: 1. clearing, 2. typing.

Naturally we still want to do these two operations underneath, but let's look for a consistent, higher-level description of that. Why are we doing these two things? Because we want to *overwrite* the text.
So I'd rename this to `overwriteIfPresent`.

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/JiraRteEditPageTest.kt
```diff
@@ -0,0 +1,138 @@
...
+            meter = meter,
+            issueMemory = issueMemory
+        )
+        val addComment = AddCommentAction(
+            jira = jira,
+            meter = meter,
+            issueMemory = issueMemory
+        )
+
+        val scenario: MutableList<Action> = mutableListOf(
```
dagguh: Do you use IntelliJ IDEA? If so, note that it will underline `scenario` in yellow, suggesting to  inline it. There's many other useful code hygiene hints, e.g. over the unnecessary `constructor()`.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/RichTextEditorTextArea.kt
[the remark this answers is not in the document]
kema2000: oh, got it, will do

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/RichTextEditorTextArea.kt
```diff
@@ -0,0 +1,59 @@
...
+            driver.wait(timeout = Duration.ofSeconds(2), condition = elementToBeClickable(webElement))
+            webElement.clear()
+            webElement.sendKeys(text)
+        }
+    }
+
+    fun clearAndTypeIfPresentRte(id : String, text: String) {
+        //check richeditor-cover exists in class attribute
+        if (driver.isElementPresent(By.id(id))) {
+            val webElement = driver.findElement(By.id(id))
```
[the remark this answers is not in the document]
kema2000: yes, agree
I removed the clearAndTypeIfPresent method, initially I was planning to re-use the existing method, but it's not required any more

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/RichTextEditorTextArea.kt
kema2000: added some comments

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/RichTextEditorTextArea.kt
kema2000: find by iframe will return more than one, e.g. Environment, Description, Comment etc.
We need to find the one associated with the target textarea

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/RichTextEditorTextArea.kt
kema2000: yep, that's correct, was mainly for debug purpose during dev

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/JiraRteEditPageTest.kt
kema2000: good to know :) thanks

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/JiraRteEditPageTest.kt
kema2000: I will increase it by 4 mins then :)

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/RichTextEditorTextArea.kt
dagguh: That's why I suggested to call it on the `webElement` instead of `driver`

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/RichTextEditorTextArea.kt
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    private fun overwritePlain(webElement: WebElement, text: String) {
+        driver.wait(timeout = Duration.ofSeconds(2), condition = elementToBeClickable(webElement))
+        webElement.clear()
+        webElement.sendKeys(text)
+    }
+
+    private fun overwriteRich(id: String, text: String) {
+        val iframeXpath = "//div[textarea[@id='$id']]//iframe"
```
dagguh: If we avoid re-finding the textarea and focus only on digging out the underlying iframe, then we can stop asking for `id: String` and expect an `textarea: WebElement` from the very top. That also means that `RichTextEditorTextArea` is no longer responsible for finding it and therefore doesn't have to handle optionality. If it's the `EditIssuePage`, which has the knowledge that the `description` textarea is optional, then let them make the call. Let them detect the presence and handle optionality.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/RichTextEditorTextArea.kt
```diff
@@ -0,0 +1,65 @@
...
+        if (optional && !driver.isElementPresent(By.id(id))) return
+
+        val wait = WebDriverWait(driver, 2)
+        val webElement = wait.until { driver -> driver.findElement(By.id(id)) }
+        val attrClass = webElement.getAttribute("class")
+        if(attrClass != null && attrClass.contains(other = "richeditor-cover", ignoreCase = true)){
+            overwriteRich(id, text)
+        }
+        else{
+            //try plain text editing
```
dagguh: This inline comment does not say more than the code itself

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/RichTextEditorTextArea.kt
dagguh: PS. Even better, we can then inject `textarea: WebElement` directly into the constructor.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/RichTextEditorTextArea.kt
```diff
@@ -0,0 +1,65 @@
...
+        driver.wait(timeout = Duration.ofSeconds(10), condition = frameToBeAvailableAndSwitchToIt(By.xpath(iframeXpath)))
+
+        //tinymce uses body element with id 'tinymce' to hold the text, find the body element
+        val bodyElement = driver.findElement(By.id("tinymce"))
+        //data-projectkey={project}
+        driver.wait(timeout = Duration.ofSeconds(10), condition = attributeToBeNotEmpty(bodyElement, "data-projectkey"))
+        bodyElement.clear()
+        bodyElement.sendKeys(text)
+
+        driver.switchTo().parentFrame()
```
dagguh: If there's an exception before this line, we will not return to the parent frame.

```kotlin
try {
    driver.wait(
        condition = frameToBeAvailableAndSwitchToIt(By.xpath(iframeXpath)),
        timeout = Duration.ofSeconds(10)
    )
    val tinyMce = driver.findElement(By.id("tinymce"))
    driver.wait(
        condition = attributeToBeNotEmpty(tinyMce, "data-projectkey"),
        timeout = Duration.ofSeconds(10)
    )
    tinyMce.clear()
    tinyMce.sendKeys(text)
} finally {
    driver.switchTo().parentFrame()
}
```
This makes sure we always switch to the parent frame, regardless of exceptions within the `try`.
Not that I also removed the inline comments, because they said literally the same words as the code already did.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/RichTextEditorTextArea.kt
```diff
@@ -0,0 +1,65 @@
...
+import org.openqa.selenium.WebElement
+import org.openqa.selenium.support.ui.ExpectedConditions
+import org.openqa.selenium.support.ui.ExpectedConditions.*
+import java.time.Duration
+import org.openqa.selenium.support.ui.WebDriverWait
+
+
+
+/**
+ * Helper class for Rich Text Editor enabled Textarea
```
dagguh: Documentation should explain what does the class represent or do.
"Helper" doesn't mean much (same as "util", "bean", "thing", "object").
And we already know it's a class, the code is right there and it is also not the most interesting aspect.

IMHO the most useful aspects are:
* represents the [Jira Rich Text Editor]( https://confluence.atlassian.com/adminjiraserver/rich-text-editing-938847886.html)
* it falls back to the plain text editor if RTE is disabled

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/RichTextEditorTextArea.kt
dagguh: I recommend using the IntelliJ debugger. It a lot of power and flexibility in comparison to temporary logging. And it doesn't risk merging leftovers :)

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/JiraRteEditPageTest.kt
dagguh: You can do it within this PR itself. Note the link in my comment.

kema2000: sure

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/RichTextEditorTextArea.kt
kema2000: sorry, I did make it clear.
The iFrame is not a sub element of Textarea field, the iFrame and the associated Textarea are sharing a same Div as parent element. We may could obtain the parent Div element, but I think there is not much difference compare to just call driver.find.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/RichTextEditorTextArea.kt
kema2000: yep, that will be better, will do

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/RichTextEditorTextArea.kt
kema2000: good point, will do

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/RichTextEditorTextArea.kt
```diff
@@ -0,0 +1,74 @@
...
+
+    private fun overwritePlain(text: String) {
+        driver.wait(timeout = Duration.ofSeconds(2), condition = elementToBeClickable(textArea))
+        textArea.clear()
+        textArea.sendKeys(text)
+    }
+
+    private fun overwriteRich(text: String) {
+        /**
+         * the tinymce Iframe element structure:
```
dagguh: This clears things up very well, nice! :)

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/CommentForm.kt
```diff
@@ -34,7 +34,8 @@ class CommentForm(
     fun enterCommentText(
         comment: String
     ): CommentForm {
-        driver.findElement(By.id("comment")).sendKeys(comment)
+        RichTextEditorTextArea(driver, driver.findElement(By.id("comment")))
```
dagguh: This is a very interesting change from the perspective of the lib clients. Let's document it in the `CHANGELOG.md`.
IMHO we should promote this change to a behavioral contract, ie. we:
```
### Added
- Support RTE with a fallback to plain-text editor. Resolve [JPERF-XXX].
```
This will mean the feature will be available in the next minor release (3.4.0). Thanks to SemVer, it will communicate to clients that they can use 3.4.0 and higher to rely on this feature and it will not be silently removed. They'll be able to remove the RTE config from the setup phase without fear.

wyrzyk:
@kema2000 Please make sure the change works well with Jira 7.2-8.0. We had a couple of attempts to support RTE in JPT, but RTE has changed multiple times between Jira versions and every time some versions were failing.

wyrzyk:
@kema2000 Please squash the commits and use issue key in commit messages (from the public issue tracker).

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/CommentForm.kt
kema2000: sure

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/RichTextEditorIT.kt
```diff
@@ -0,0 +1,163 @@
...
+/**
+ * Test Editing issue without SetupAction (i.e Rich Text editor enabled)
+ * Covers:
+ *      EditIssue
+ *      AddComment
+ */
+class RichTextEditorIT {
+
+    @Test
+    fun shouldRunScenarioWithoutErrors() {
```
wyrzyk: Test code is not only about verifying the state of the application. It's also living documentation (IMO the best one because it can't be outdated). Please consider making the test more readable. 
See: http://wiki.c2.com/?ArrangeActAssert

wyrzyk: I see `runTestOnOneInstance` is hidden in runTestOnOneInstance and the test isn't a test, it runs a suite of tests. Consider ignoring my previous comment.

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/RichTextEditorIT.kt
```diff
@@ -0,0 +1,163 @@
...
+        )
+        var basePort = 8080;
+
+        for (version in versions) {
+            runTestOnOneInstance(version, basePort++)
+        }
+    }
+
+    private fun runTestOnOneInstance(version: String, port : Int) {
+        println("Testing JIRA with port : $port, version : $version")
```
wyrzyk: Why not log?

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/RichTextEditorIT.kt
```diff
@@ -0,0 +1,163 @@
...
+        val versions = listOf(
+            "8.0.0"
+//            "7.13.0"
+//            "7.3.0"
+//            "7.2.0"
+        )
+        var basePort = 8080;
+
+        for (version in versions) {
+            runTestOnOneInstance(version, basePort++)
```
wyrzyk: I wish we could run it in parallel.

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/RichTextEditorIT.kt
```diff
@@ -0,0 +1,163 @@
...
+                        SeededRandom(123),
+                        actionMeter
+                    )
+
+                    logInAction.run()
+                    actions.forEach { action ->
+                        action.run()
+                    }
+                }
+                jira.close()
```
wyrzyk: You don't have to invoke .close() here. `use` will close the resource.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/EditIssuePage.kt
```diff
@@ -31,7 +30,11 @@ class EditIssuePage(
 
     fun fillForm(): EditIssuePage {
         summaryLocator.clearAndTypeIfPresent("summary")
-        descriptionLocator.clearAndTypeIfPresent("description")
+        val descriptionFieldId = "description"
+        if(driver.isElementPresent(By.id(descriptionFieldId))){
+            RichTextEditorTextArea(driver, driver.findElement(By.id(descriptionFieldId)))
+                .overwriteIfPresent("description")
```
wyrzyk: Is the "description" here the same as `descriptionFieldId`. If not, will it work when Jira's language is other than English?

wyrzyk:
@kema2000 @dagguh The PR seems to stick around for a while. How can I help you to move it forward?

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/EditIssuePage.kt
kema2000: It's the value to be sent to Description field. In general, English text shall be accepted regardless the language set.

kema2000:
> @kema2000 @dagguh The PR seems to stick around for a while. How can I help you to move it forward?

I think the main code is ready to be merged. I am planning to enhance the UT (to test multiple jira version in parallel). But that requires quite bit code change to docker part.

wyrzyk:
Nothing happens here for more than a week. I've created a PR based on this one that addresses the feedback (https://github.com/atlassian/jira-actions/pull/8).

wyrzyk:
The changes have been merged. See https://github.com/atlassian/jira-actions/pull/8.

PR #6, opened 2019-04-01
Issue/quick 105

kema2000:
I made WebJira.adminPassword public, so during HWR testing, we don't need to hard-code the password for the Admin Access dialog

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ stages:
 jobs:
   include:
     - stage: test
-      script: travis_wait 14 ./gradlew build
+      script: travis_wait 21 ./gradlew build
     - stage: release
       script: ./scripts/travis-release.sh
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,11 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://bitbucket.org/atlassian/jira-actions/branches/compare/master%0Drelease-3.3.0
 
+### Fixed - 2019-02-27
+- Support RTE with a fallback to plain-text editor. Resolve [JPERF-184]
+
+[JPERF-184]: https://ecosystem.atlassian.net/browse/JPERF-184
+
 ## [3.3.0] - 2019-01-29
 [3.3.0]: https://bitbucket.org/atlassian/jira-actions/branches/compare/release-3.3.0%0Drelease-3.2.0
 
--- a/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/WebJira.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/WebJira.kt
@@ -84,4 +84,9 @@ data class WebJira(
     fun navigateTo(path: String) {
         driver.navigate().to(base.resolve(path).toURL())
     }
+
+    fun adminPassword(): String {
+        return adminPassword
+    }
+
 }
--- a/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/WebJira.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/WebJira.kt
@@ -9,7 +9,7 @@ import java.net.URLEncoder
 data class WebJira(
     val driver: WebDriver,
     val base: URI,
-    private val adminPassword: String
+    val adminPassword: String
 ) {
 
     fun goToLogin(): LoginPage {
--- a/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/WebJira.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/WebJira.kt
@@ -9,7 +9,7 @@ import java.net.URLEncoder
 data class WebJira(
     val driver: WebDriver,
     val base: URI,
-    val adminPassword: String
+    private val adminPassword: String
 ) {
 
     fun goToLogin(): LoginPage {
--- a/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/EditIssuePage.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/EditIssuePage.kt
@@ -3,6 +3,7 @@ package com.atlassian.performance.tools.jiraactions.api.page
 import com.atlassian.performance.tools.jiraactions.page.form.IssueForm
 import org.openqa.selenium.*
 import org.openqa.selenium.support.ui.ExpectedConditions.*
+import org.openqa.selenium.support.ui.Select
 import java.time.Duration
 
 
@@ -36,7 +37,19 @@ class EditIssuePage(
                 .overwriteIfPresent("description")
         }
         form.fillRequiredFields()
+        selectResolution()
         return this
+    }
+
+    private fun selectResolution() {
+        val resolutionLocator = By.id("resolution")
+        if (driver.isElementPresent(resolutionLocator)) {
+            val dropDown = Select(driver.findElement(resolutionLocator))
+            if(dropDown.options != null && dropDown.options.size > 1){
+                val selection = dropDown.options.get(1)
+                dropDown.selectByVisibleText(selection.text)
+            }
+        }
     }
 
     fun submit(): IssuePage {
--- a/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/EditIssuePage.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/EditIssuePage.kt
@@ -1,17 +1,16 @@
 package com.atlassian.performance.tools.jiraactions.api.page
 
 import com.atlassian.performance.tools.jiraactions.page.form.IssueForm
-import org.openqa.selenium.By
-import org.openqa.selenium.WebDriver
+import org.openqa.selenium.*
 import org.openqa.selenium.support.ui.ExpectedConditions.*
 import java.time.Duration
+
 
 class EditIssuePage(
     private val driver: WebDriver
 ) {
     private val form = IssueForm(By.id("issue-edit"), driver)
     private val summaryLocator = By.id("summary")
-    private val descriptionLocator = By.id("description")
     private val updateButtonLocator = By.id("issue-edit-submit")
 
     fun waitForEditIssueForm(): EditIssuePage {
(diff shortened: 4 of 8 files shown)


src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/WebJira.kt
```diff
@@ -9,7 +9,7 @@ import java.net.URLEncoder
 data class WebJira(
     val driver: WebDriver,
     val base: URI,
-    private val adminPassword: String
+    val adminPassword: String
```
wyrzyk: Adding a new property to the data class is an API breaking change. Please see https://hello.atlassian.net/wiki/spaces/JIRASERVER/blog/2018/12/18/370711047/Is+Kotlin+a+good+choice+for+a+Java+library#Kotlin-data-classes-are-a-poor-choice-for-API.

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/RichTextEditorIT.kt
```diff
@@ -0,0 +1,163 @@
...
+         *
+         * This is not working now due to:
+         * 1. the static network alias is used
+         * 2. the jira.close() does not release the resource properly yet
+         *
+         * So we will keep it for one version per run now.
+         */
+        val versions = listOf(
+            "8.0.0"
+//            "7.13.0"
```
wyrzyk: It would be awesome to keep testing multiple Jira versions.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/WebJira.kt
dagguh: Publishing and existing field does not break ABI.
If you decompile the current `WebJira.kt` to Java, you'll see:
```
   @NotNull
   public final WebDriver component1() {
      return this.driver;
   }

   @NotNull
   public final URI component2() {
      return this.base;
   }

   private final String component3() {
      return this.adminPassword;
   }
```

```
   @NotNull
   public final WebJira copy(@NotNull WebDriver driver, @NotNull URI base, @NotNull String adminPassword) {
      Intrinsics.checkParameterIsNotNull(driver, "driver");
      Intrinsics.checkParameterIsNotNull(base, "base");
      Intrinsics.checkParameterIsNotNull(adminPassword, "adminPassword");
      return new WebJira(driver, base, adminPassword);
   }
```

```
   // $FF: synthetic method
   @NotNull
   public static WebJira copy$default(WebJira var0, WebDriver var1, URI var2, String var3, int var4, Object var5) {
      if ((var4 & 1) != 0) {
         var1 = var0.driver;
      }

      if ((var4 & 2) != 0) {
         var2 = var0.base;
      }

      if ((var4 & 4) != 0) {
         var3 = var0.adminPassword;
      }

      return var0.copy(var1, var2, var3);
   }
```

The private field is already handled.

CHANGELOG.md
```diff
@@ -26,6 +26,11 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://bitbucket.org/atlassian/jira-actions/branches/compare/master%0Drelease-3.3.0
 
+### Fixed - 2019-02-27
+- Support RTE with a fallback to plain-text editor. Resolve [JPERF-184]
+
+[JPERF-184]: https://ecosystem.atlassian.net/browse/JPERF-184
```
dagguh: This is a resubmitted https://github.com/atlassian/jira-actions/pull/3 with some extra changes on top. Should we close that one?

.travis.yml
```diff
@@ -8,7 +8,7 @@ stages:
 jobs:
   include:
     - stage: test
-      script: travis_wait 14 ./gradlew build
+      script: travis_wait 21 ./gradlew build
```
dagguh: Extending the dev loop by 7 minutes is a lot.
`virtual-users` have a Docker Jira with 100k users, which starts in 7 minutes (with 1 minute upgrade task time and no jira-home/DB parallelization). `JiraCoreFormula` might be a bit too slow. We might leverage some form of caching (e.g. the emergent dataset).
The dev loop penalty is induced by the inefficiencies in `docker-infrastructure`, so I won't require this to be fixed here. However, if we do expand the coverage over multiple Jira versions, I'd ask to parallelize the test internally to keep the dev loop penalty reasonably low.

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/RichTextEditorIT.kt
```diff
@@ -0,0 +1,163 @@
...
+
+        return mutableListOf(
+            createIssue,
+            searchWithJql,
+            viewIssue,
+            editIssue,
+            addComment
+        )
+    }
+
```
sebapawlak: Double empty lines

CHANGELOG.md
```diff
@@ -26,6 +26,11 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://bitbucket.org/atlassian/jira-actions/branches/compare/master%0Drelease-3.3.0
 
+### Fixed - 2019-02-27
```
sebapawlak: Shouldn't it actually be put in `Added` section? We are adding new API which in this PR as far as I can see.

dagguh: `RichTextEditorTextArea` is new API, so we definitely `Added` contracts

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/EditIssuePage.kt
```diff
@@ -36,9 +37,21 @@ class EditIssuePage(
                 .overwriteIfPresent("description")
         }
         form.fillRequiredFields()
+        selectResolution()
         return this
     }
 
+    private fun selectResolution() {
```
dagguh: You're mixing 3 different features in the same PR. It means that if there's a problem with any of them, all get blocked. It makes merging and reviewing harder. I recommend splitting PRs per feature.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/page/IssueCreateDialog.kt
```diff
@@ -49,12 +54,32 @@ internal class IssueCreateDialog(
...
+
+        driver.wait(elementToBeClickable(configColumnField)).click()
+        val locator = By.xpath("//div[@class='qf-picker-header']/dl/dd/a[text()='All']")
+        if (driver.isElementPresent(locator)) {
+            driver.wait(elementToBeClickable(locator)).click()
+        } else {
+            driver.wait(elementToBeClickable(configColumnField)).click()
+        }
         form.fillRequiredFields()
+        selectResolution()
```
dagguh: Why doesn't the `form.fillRequiredFields()` cover that already? Is the resolution required? If so, why didn't we detect it? It'd be much easier to reason about if we had a red test that shows the need for this change.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/EditIssuePage.kt
kema2000: My mistake, I meant to just push the changes so everyone in the team can view/edit it. Did not realise that I raised a PR earlier.

kema2000:
Given we are using this branch for XL HWR, and we are still making changes quite often. Feel free to leave comments here, I will address them after all the code base is stabilized.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/EditIssuePage.kt
dagguh: You can create a new branch for the purposes of getting XL results. Pushing many independent changes via a single branch won't work.

Do you want to merge these changes? If so, split them. If not, then let's close this PR, because reviewers are spending energy on the reviews.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/page/IssueCreateDialog.kt
```diff
@@ -49,12 +54,32 @@ internal class IssueCreateDialog(
     }
 
     fun fillRequiredFields(): IssueCreateDialog {
+
+        driver.wait(elementToBeClickable(configColumnField)).click()
+        val locator = By.xpath("//div[@class='qf-picker-header']/dl/dd/a[text()='All']")
```
dagguh: So this shows all issue fields in the dialog.
AFAIK the HWR XL7 dataset is problematic, because some projects/issue types have a mandatory `Resolution` field, which has no default value *and* it's not visible on the screens by default.

dagguh: The selector is way too specific. Do we really care about `div`s, `dl`s and `dd`s?
These page objects are supposed to work with multiple Jira versions. Let's pick the most semantic selector, which won't break after some DOM changes. Ie. let's look for the element with `Configure Fields` text and then the `All` text within it.

kema2000: Agree, the root cause is the dataset. Making all field available might add values for people using their own dataset.
Do we think we might need this in the future? If so I will create a new branch to pick those changes and raise a new PR.

kema2000: in the dom tree, the "All" element is not contained within "Configure Fields" element.
Maybe I can simplify it as:
//*[@id='inline-dialog-field_picker_popup']//a[text()='All']

dagguh: Looks good 👍

dagguh: Let's raise a new PR. Let's include the context, the purpose of this change, because beforehand it wasn't clear. Ideally we'd write an integration test with such a dataset (funky issue screen, not necessarily XL), but if we need to cut scope, then at least let's explain it in the commit message or method Javadoc.

kema2000: sure, will do

PR #7, opened 2019-04-05
Expose `IssueForm` to make custom issue form actions easier

dagguh:
We're having some problems with our dataset. There's some problem with the "resolution" field.
We're exploring some workarounds, e.g.: https://github.com/kema2000/jira-actions/commit/72eefabbd15308fbf8c701693dba7f5079b7a564
However, we're forced to work on lib snapshots. I'd rather work on my custom `EditIssueAction`, which needs `EditIssuePage`, which needs `IssueForm`. The first two are very case-specific and can differ from dataset to dataset, so I don't expect them to be changed in `jira-actions`. However, copy-pasting `IssueForm` (along with all transitive internals) seems to be coutner-productive, because it contains pretty universal logic.
Therefore I want to expose it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,11 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/jira-actions/compare/release-3.4.0...master
 
+### Added
+- Expose `IssueForm` to make custom issue form actions easier. Resolve [JPERF-450].
+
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -26,6 +26,9 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/jira-actions/compare/release-3.4.0...master
 
+### Added
+- Expose `IssueForm` to make custom issue form actions easier.
```
wyrzyk: Missing issue key.

PR #8, opened 2019-04-24
JPERF-184 Add support for Rich Text Editor

wyrzyk:
Source: https://github.com/atlassian/jira-actions/pull/3

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ stages:
 jobs:
   include:
     - stage: test
-      script: travis_wait 14 ./gradlew build
+      script: travis_wait 21 ./gradlew build
     - stage: release
       script: ./scripts/travis-release.sh
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,10 @@
 language: java
 
+env:
+  - JIRA_SOFTWARE_VERSION=8.0.0
+  - JIRA_SOFTWARE_VERSION=7.13.0
+  - JIRA_SOFTWARE_VERSION=7.3.0
+  - JIRA_SOFTWARE_VERSION=7.2.0
 stages:
   - test
   - name: release
@@ -10,6 +15,7 @@ jobs:
     - stage: test
       script: travis_wait 21 ./gradlew build
     - stage: release
+      env: JIRA_SOFTWARE_VERSION=none
       script: ./scripts/travis-release.sh
 
 before_cache:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,10 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ### Added
 - Expose `IssueForm` to make custom issue form actions easier. Resolve [JPERF-450].
 
+### Fixed
+- Support RTE with a fallback to plain-text editor. Resolve [JPERF-184].
+
+[JPERF-184]: https://ecosystem.atlassian.net/browse/JPERF-184
 [JPERF-450]: https://ecosystem.atlassian.net/browse/JPERF-450
 
 ## [3.4.0] - 2019-03-28
--- a/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/EditIssuePage.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/EditIssuePage.kt
@@ -11,8 +11,8 @@ class EditIssuePage(
 ) {
     private val form = IssueForm(By.id("issue-edit"), driver)
     private val summaryLocator = By.id("summary")
-    private val descriptionLocator = By.id("description")
     private val updateButtonLocator = By.id("issue-edit-submit")
+    private val descriptionFieldLocator = By.id("description")
 
     fun waitForEditIssueForm(): EditIssuePage {
         val jiraErrors = JiraErrors(driver)
@@ -31,7 +31,10 @@ class EditIssuePage(
 
     fun fillForm(): EditIssuePage {
         summaryLocator.clearAndTypeIfPresent("summary")
-        descriptionLocator.clearAndTypeIfPresent("description")
+        if (driver.isElementPresent(descriptionFieldLocator)) {
+            RichTextEditorTextArea(driver, driver.findElement(descriptionFieldLocator))
+                .overwriteIfPresent("description")
+        }
         form.fillRequiredFields()
         return this
     }
--- a/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/RichTextEditorTextArea.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/RichTextEditorTextArea.kt
@@ -1,0 +1,64 @@
+package com.atlassian.performance.tools.jiraactions.api.page
+
+import org.openqa.selenium.By
+import org.openqa.selenium.WebDriver
+import org.openqa.selenium.WebElement
+import org.openqa.selenium.support.ui.ExpectedConditions.*
+import java.time.Duration
+
+/**
+ * Represents the [Jira Rich Text Editor](https://confluence.atlassian.com/adminjiraserver/rich-text-editing-938847886.html)
+ * falls back to the plain text editor if RTE is disabled.
+ */
+class RichTextEditorTextArea(
+    private val driver: WebDriver,
+    private val textArea: WebElement
+) {
+
+    fun overwriteIfPresent(text: String) {
+        val attrClass = textArea.getAttribute("class")
+        if(attrClass != null && attrClass.contains(other = "richeditor-cover", ignoreCase = true)){
+            overwriteRich(text)
+        }
+        else{
+            overwritePlain(text)
+        }
+    }
+
+    private fun overwritePlain(text: String) {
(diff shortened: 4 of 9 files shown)


src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/RichTextEditorIT.kt
```diff
@@ -0,0 +1,144 @@
...
+ *      EditIssue
+ *      AddComment
+ */
+class RichTextEditorIT {
+    private val logger: Logger = LogManager.getLogger(this::class.java)
+
+    @Test
+    fun shouldRunScenarioWithoutErrors() {
+        val port = 8080
+        val version = "8.0.0"
```
dagguh: So we're sticking with 8.0.0 only. That's fine, but I want to make sure it's deliberate.

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/RichTextEditorIT.kt
```diff
@@ -0,0 +1,144 @@
...
+        logger.info("Testing JIRA with port : $port, version : $version")
```
dagguh: `JIRA` -> `Jira`
Also, how does this logging help understand the test?

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/EditIssuePage.kt
```diff
@@ -31,7 +30,11 @@ class EditIssuePage(
 
     fun fillForm(): EditIssuePage {
         summaryLocator.clearAndTypeIfPresent("summary")
-        descriptionLocator.clearAndTypeIfPresent("description")
+        val descriptionFieldId = "description"
+        if(driver.isElementPresent(By.id(descriptionFieldId))){
```
dagguh: Let's reformat, the line is very cramped.
PS. `By.id(descriptionFieldId)` is extractable.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/RichTextEditorTextArea.kt
```diff
@@ -0,0 +1,66 @@
...
+         *           --iframe
+         */
+        val iframeXpath = "//div[textarea[@id='${textArea.getAttribute("id")}']]//iframe"
+
+        try {
+            driver.wait(
+                timeout = Duration.ofSeconds(10),
+                condition = frameToBeAvailableAndSwitchToIt(By.xpath(iframeXpath))
+            )
+            //tinymce uses body element with id 'tinymce' to hold the text, find the body element
```
dagguh: This comment doesn't provide any additional context WRT the line below.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/RichTextEditorTextArea.kt
```diff
@@ -0,0 +1,66 @@
...
+            val tinyMce = driver.findElement(By.id("tinymce"))
+            //data-projectkey={project}
```
dagguh: What does it mean?

wyrzyk: Have not idea

CHANGELOG.md
```diff
@@ -29,6 +29,11 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ### Added
 - Expose `IssueForm` to make custom issue form actions easier. Resolve [JPERF-450].
 
+### Fixed
+- Support RTE with a fallback to plain-text editor. Resolve [JPERF-184]
+
+[JPERF-184]: https://ecosystem.atlassian.net/browse/JPERF-184
+
```
mcmurphy2000: Should we remove this extra line?

CHANGELOG.md
```diff
@@ -29,6 +29,11 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ### Added
 - Expose `IssueForm` to make custom issue form actions easier. Resolve [JPERF-450].
 
+### Fixed
+- Support RTE with a fallback to plain-text editor. Resolve [JPERF-184]
```
mcmurphy2000: Dot at the end?

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/EditIssuePage.kt
mcmurphy2000: Why did you stop putting a space before `{` ? ;)

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/RichTextEditorIT.kt
wyrzyk: Yes. I just want to merge it and reduce our work in progress. I don't have the capacity to make the test parallel and I don't want to make our dev loop slower (by running the test in a sequence).

dagguh: *If* we're skipping Jira 7.2.0/7.3.0 (see `RichTextEditorConfiguration` comments), then it's a bit risky. Note that this change is not opt-in and there's no easy opt-out (apart from never bumping). This PR changes how the `EditIssuePage` and `CommentForm` work and if the RTE detection doesn't work on 7.2.0/7.3.0, we're breaking them.

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/RichTextEditorIT.kt
wyrzyk: It doesnt

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/RichTextEditorIT.kt
wyrzyk: You're right. I'm not sure if it's worth pulling the change.

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/RichTextEditorIT.kt
wyrzyk: Removed

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/RichTextEditorIT.kt
dagguh: If can we cheaply run it for 7.2.0, 7.3.0 and 8.0.0 in parallel, then it's worth it IMHO.

wyrzyk: I've got another idea, but it could be controversial. We could choose the version randomly. None of them will be favoured. We can also schedule daily build to see if it flakes on some versions.

The problem of compatibility with a wide range of versions is not a problem of this test. The whole module should be tested with different Jira versions.

sebapawlak: I'd rather test 8.0.0 only than test random versions.

dagguh: Agreed with @sebapawlak 
Randomness can lead to us merging a bug that manifests itself on 7.3.0 and if the PR CI skips over it randomly, we'll get a fail late, perhaps even after a release.
IMHO parallielization complexity is worth it and we've been doing it a lot already, it works.

> The whole module should be tested with different Jira versions.

Maybe. But we can do it without randomness. E.g. we can add Gradle tasks per Jira version, which would set a system property, read by the tests (with a default).
However, we don't need to do it in this PR. It's big enough already.

wyrzyk:
I don't understand why Travis runs 8.0.0 twice.

.travis.yml
```diff
@@ -8,7 +8,10 @@ stages:
 jobs:
   include:
     - stage: test
-      script: travis_wait 14 ./gradlew build
+      script: travis_wait 21 ./gradlew build
+    - script: travis_wait 21 ./gradlew -Djira.version=7.13.0 build
+    - script: travis_wait 21 ./gradlew -Djira.version=7.3.0 build
+    - script: travis_wait 21 ./gradlew -Djira.version=7.2.0 build
```
sebapawlak: Are those run in parallel?

wyrzyk: Yes, but some jobs are queued and start later due to lack of agents (AFAIK there's a limited OS agents pool).

wyrzyk: I've switched back to the matrix, as it creates more readable jobs. I still don't know, why it runs 8.0.0 twice.

dagguh: Thank god we can cover 4 versions of Jira here and we **do not** have to do it for all other tests in all the other modules.

PR #10, opened 2019-05-14
Observations added to View Issue action.

drspartez:
Included observations enable us to match performance results and underlying data.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,11 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/jira-actions/compare/release-3.5.0...master
 
+### Fixed
+- Observe issue key and id in ViewIssueAction. Resolves [JPERF-482].
+
+[JPERF-482]: https://ecosystem.atlassian.net/browse/JPERF-482
+
 ## [3.5.0] - 2019-04-26
 [3.4.0]: https://github.com/atlassian/jira-actions/compare/release-3.4.0...release-3.5.0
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -26,6 +26,9 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/jira-actions/compare/release-3.5.0...master
 
+### Added
```
dagguh: If we want to *add* this as a new contract, we should attach it to the public javadoc, e.g. in `ViewIssueAction`, describing the promised behavior.

CHANGELOG.md
```diff
@@ -26,6 +26,9 @@ Dropping a requirement of a major version of a dependency is a new contract.
...
+- View Issue action updated to provide observation data together with metric (issueKey and issueId).
```
dagguh: Less boilerplate:
> Observe issue key and id in `ViewIssueAction`.

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/JiraCoreScenarioIT.kt
```diff
@@ -82,5 +83,10 @@ class JiraCoreScenarioIT {
             metric.result
         }
         Assertions.assertThat(results).containsOnly(ActionResult.OK)
+
```
dagguh: This blank line disrupts the implicit given-when-then test stages

CHANGELOG.md
sebapawlak: AFAIK we don't want it as a _contract_. Let's delete this entry @drspartez

dagguh: If we don't consider it a behavioral contract then we can remove it at any point. It also means that the users can use these new observations *if* they're present, but they might go missing at any point.
We mark these as `### Fixed`.

PR #15, opened 2019-07-08
Mark 3.6.0

dagguh:


dagguh:
Thanks for the ping Paweł
Fixed

PR #16, opened 2019-07-10
RUM-2812: spike JqlMemory change to categorize queries

drspartez:
This pull request is for feedback only about API change. Do not focus on names  (for prescriptions keys) and other minor stuff.
Goal was to recall JQL from memory by some key/tag/category name

### DIFF
--- a/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/memories/JqlMemory.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/memories/JqlMemory.kt
@@ -3,5 +3,10 @@ package com.atlassian.performance.tools.jiraactions.api.memories
 import com.atlassian.performance.tools.jiraactions.api.page.IssuePage
 
 interface JqlMemory : Memory<String> {
+
     fun observe(issuePage: IssuePage)
+
+    fun recall(filter: (String) -> Boolean): String? {
+        return null;
+    }
 }
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -47,6 +47,7 @@ dependencies {
     testCompile("com.atlassian.performance.tools:io:[1.0.0,2.0.0)")
     testCompile("com.atlassian.performance.tools:docker-infrastructure:0.1.2")
     testCompile("junit:junit:4.12")
+    testCompile("org.mockito:mockito-core:2.16.0")
 }
 
 tasks.test {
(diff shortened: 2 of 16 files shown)


src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/memories/JqlMemory.kt
```diff
@@ -3,5 +3,10 @@ package com.atlassian.performance.tools.jiraactions.api.memories
 import com.atlassian.performance.tools.jiraactions.api.page.IssuePage
 
 interface JqlMemory : Memory<String> {
+
     fun observe(issuePage: IssuePage)
+
+    fun recall(filter: (String) -> Boolean): String? {
```
wyrzyk: 1. It's hard to use it without documentation. We don't know what the String represents. Can we change it to a class/enum?

lipinskipawel: I would go with enum. 
1. The client of this code can easily query through all possibilities.
2. All possibilities are predefined (this is not a cache, we should avoid putting any object into it)

drspartez: I though about an enum also. Why i used string:
1) I'd like to avoid adding next API artifact to this code
2) This enum would be a redundant type to JqlPerscriptions which in fact somehow pretend enumeration - but with minor change it is not possible to refactor 
3) Any new built-in JQL provided in the same fashion would require change in JqlPrescriptions, AdaptiveJqlMemory and the enum itself

There is one more thing, JqlMemory may be fed by user JQL's, and then it would be nice to provide such "tagging" for user queries.
We easily can achieve that by providing **remember(memories: Collection<BakedSql>)**
to JqlMemory (BakedSql is API already).
If we go for enum we need to provide the way to extend/implement user defined classification (interface on enum ??? :)).
 
Due to KISS rule I did string.
WDYT ?

lipinskipawel: As a client of the JqlMemory I would like to specify which type of JQL I want.
Maybe we should think about something like that:
`fun recall(obj : SomeJqlEnum) : List<String>`
Why we shouldn't pass 

> (String) -> Boolean

, this will means that we expose some internal logic to the client. 
He can use our API in an improper way (AdaptiveJqlMemory is not just a list, map of some items).

wyrzyk: @drspartez Let's consider how we could evolve API based on String and Object/Enum in the future. Which implementation will be more flexible in case we decide to filter JQL's by a different factor?

PR #17, opened 2019-07-11
JPERF-522: Added recallByTag method JqlMemory (default) and implementation

drspartez:
Initial discussion was here https://github.com/atlassian/jira-actions/pull/16
This is not invasive change that provides ability to create SearchJql actions  with user selected JQLS

### DIFF
--- a/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/memories/JqlMemory.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/memories/JqlMemory.kt
@@ -1,7 +1,14 @@
 package com.atlassian.performance.tools.jiraactions.api.memories
 
 import com.atlassian.performance.tools.jiraactions.api.page.IssuePage
+import java.util.function.Predicate
 
 interface JqlMemory : Memory<String> {
+
     fun observe(issuePage: IssuePage)
+
+    @JvmDefault
+    fun recallByTag(filter: Predicate<String>): String? {
+        return null
+    }
 }
--- a/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/memories/adaptive/AdaptiveJqlMemory.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/memories/adaptive/AdaptiveJqlMemory.kt
@@ -5,8 +5,10 @@ import com.atlassian.performance.tools.jiraactions.api.memories.JqlMemory
 import com.atlassian.performance.tools.jiraactions.api.memories.adaptive.jql.JqlPrescription
 import com.atlassian.performance.tools.jiraactions.api.memories.adaptive.jql.JqlPrescriptions
 import com.atlassian.performance.tools.jiraactions.api.page.IssuePage
+import com.atlassian.performance.tools.jiraactions.jql.BuiltInJQL
 import org.apache.logging.log4j.LogManager
 import org.apache.logging.log4j.Logger
+import java.util.function.Predicate
 
 class AdaptiveJqlMemory(
     private val random: SeededRandom
@@ -14,39 +16,50 @@ class AdaptiveJqlMemory(
 
     private val logger: Logger = LogManager.getLogger(this::class.java)
 
+    companion object {
+        fun getAvailableTags(): List<String> = BuiltInJQL.values().map { it.name }.toList()
+    }
+
     private val jqls = mutableListOf(
-        "resolved is not empty order by description",
-        "text ~ \"a*\" order by summary"
+        TaggedBakedJql(BakedJql({ _ -> ""}, "resolved is not empty order by description"), BuiltInJQL.RESOLVED.name),
+        TaggedBakedJql(BakedJql({ _ -> ""}, "text ~ \"a*\" order by summary"), BuiltInJQL.GENERIC_WIDE.name)
     )
-    private val jqlPrescriptions = mutableSetOf(
-        JqlPrescriptions.prioritiesInEnumeratedList(random),
-        JqlPrescriptions.specifiedProject,
-        JqlPrescriptions.specifiedAssignee,
-        JqlPrescriptions.previousReporters,
-        JqlPrescriptions.specifiedAssigneeInSpecifiedProject,
-        JqlPrescriptions.filteredByGivenWord(random)
+
+    private val jqlPrescriptions = mutableMapOf(
+        BuiltInJQL.PRIORITIES.name to JqlPrescriptions.prioritiesInEnumeratedList(random),
+        BuiltInJQL.PROJECT.name to JqlPrescriptions.specifiedProject,
+        BuiltInJQL.ASSIGNEE.name to JqlPrescriptions.specifiedAssignee,
+        BuiltInJQL.REPORTERS.name to JqlPrescriptions.previousReporters,
+        BuiltInJQL.PROJECT_ASSIGNEE.name to JqlPrescriptions.specifiedAssigneeInSpecifiedProject,
+        BuiltInJQL.GIVEN_WORD.name to JqlPrescriptions.filteredByGivenWord(random)
     )
 
     override fun observe(issuePage: IssuePage) {
         val bakedJql = jqlPrescriptions.asSequence()
-            .map { BakedJql(it, it(issuePage)) }
-            .filter { it.jql != null }
+            .map { TaggedBakedJql(BakedJql(it.value, it.value(issuePage)), it.key) }
+            .filter { it.baked.jql != null }
             .firstOrNull()
 
         bakedJql?.let {
-            logger.debug("Rendered a new jql query: <<${it.jql!!}>>")
-            jqls.add(it.jql)
-            jqlPrescriptions.remove(it.jqlPrescription)
+            logger.debug("Rendered a new jql query: <<${it.baked.jql!!}>>")
+            jqls.add(it)
+            jqlPrescriptions.remove(it.tag)
         }
     }
 
     override fun recall(): String? {
-        return random.pick(jqls)
+        return random.pick(jqls)?.baked?.jql
     }
 
     override fun remember(memories: Collection<String>) {
-        jqls.addAll(memories)
+        jqls.addAll(memories.map { TaggedBakedJql(BakedJql({ _ -> ""}, it)) })
     }
+
+    override fun recallByTag(filter: Predicate<String>): String? {
+        return random.pick(jqls.filter {it.tag != null && filter.test(it.tag.toString()) }.toList())?.baked?.jql
+    }
+
+    private class TaggedBakedJql(val baked: BakedJql, val tag: String? = null)
 }
 
 data class BakedJql(
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,8 +28,10 @@ Dropping a requirement of a major version of a dependency is a new contract.
 
 ### Added
 - Expose `WebJira.accessAdmin()`. Resolve [JPERF-207].
+- JqlMemory modified to allow recall queries by tag/query type. Resolve [JPERF-522].
 
 [JPERF-207]: https://ecosystem.atlassian.net/browse/JPERF-207
+[JPERF-522]: https://ecosystem.atlassian.net/browse/JPERF-522
 
 ## [3.5.1] - 2019-05-22
 [3.5.1]: https://github.com/atlassian/jira-actions/compare/release-3.5.0...release-3.5.1
--- a/src/main/kotlin/com/atlassian/performance/tools/jiraactions/jql/BuiltInJQL.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/jiraactions/jql/BuiltInJQL.kt
@@ -1,0 +1,12 @@
+package com.atlassian.performance.tools.jiraactions.jql
+
+internal enum class BuiltInJQL {
+    GENERIC_WIDE,
+    RESOLVED,
+    PRIORITIES,
+    PROJECT,
+    ASSIGNEE,
+    REPORTERS,
+    PROJECT_ASSIGNEE,
+    GIVEN_WORD;
+}


src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/JqlMemoryTest.kt
```diff
@@ -0,0 +1,73 @@
...
+import org.mockito.stubbing.Answer
+import org.openqa.selenium.WebDriver
+
+class JqlMemoryTest {
+
+    @Test
+    fun allBuiltInQueryTypesAvailable() {
+        val memory: JqlMemory = AdaptiveJqlMemory(SeededRandom());
+        val webdriverMock = Mockito.mock(WebDriver::class.java)
+        val issuePageSpy = Mockito.spy(IssuePage(webdriverMock))
```
dagguh: This is unfortunately necessary if we want to mock `IssuePage`.
Ideally, we would turn `IssuePage` into an interface and use a mock implementation here, avoiding the huge runtime, dependency, readability and diagnosability cost of Mockito.

Changing a class into an interface is binary-incompatible (`invokeinterface` bytecode instruction). Well be able to clean this up only in a major release.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/memories/adaptive/AdaptiveJqlMemory.kt
```diff
@@ -14,42 +15,55 @@ class AdaptiveJqlMemory(
...
+
 }
 
 data class BakedJql(
     val jqlPrescription: JqlPrescription,
-    val jql: String?
-)
\ No newline at end of file
+    val jql: String?,
+    val tag: String?
```
dagguh: This breaks `copy` calls, because it's a `data class`. We can assume nobody calls `copy`, but we've been wrong like this before.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/memories/JqlMemory.kt
```diff
@@ -3,5 +3,10 @@ package com.atlassian.performance.tools.jiraactions.api.memories
 import com.atlassian.performance.tools.jiraactions.api.page.IssuePage
 
 interface JqlMemory : Memory<String> {
+
     fun observe(issuePage: IssuePage)
+
+    fun recallByTag(filter: (String) -> Boolean): String? {
```
dagguh: This requires `kotlin.jvm.functions.Function1`.
We can make it more portable if we switch to `java.util.function.Predicate`.

dagguh: This uses the Kotlin default implementation, which is not consumable by Java implementors, which breaks them, unless we use `@JvmDefault`.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/memories/JqlMemory.kt
```diff
@@ -3,5 +3,10 @@ package com.atlassian.performance.tools.jiraactions.api.memories
 import com.atlassian.performance.tools.jiraactions.api.page.IssuePage
 
 interface JqlMemory : Memory<String> {
```
dagguh: It seems like we want to treat JQLs not only as `String`s, but something more. We want to categorize them, tag them. We want a `Memory<Jql>`. This is why [primitive obsession](http://wiki.c2.com/?PrimitiveObsession) is a ticking time bomb. It strips us from extension points.

This is why the primitive obsession in `recallByTag` will also create problems in the future.

I propose to add a brand new interface, use domain classes (like `class Jql(val query: String, type: JqlType)`) and leave `JqlMemory` intact. The `AdaptiveJqlMemory` can implement both old and new ones at the same time.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/memories/JqlMemory.kt
drspartez: It is done the same fashion as IssueMemory recall is.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/memories/JqlMemory.kt
drspartez: I considered this, but I'm not doing refactoring this time.

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/JqlMemoryTest.kt
drspartez: I do not see reason to write more code or do refactoring when my objective different. I see mockito is widely used in atlassian/spartez projets. Because of mockito this test case is clean and neat. 
I dont't get why we would be able to clean this in major release - this is the same module.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/memories/JqlMemory.kt
wyrzyk: @drspartez IssueMemory is already part of an API. It's hard to fix it. I see no point in repeating the mistake here.

drspartez: Ok. I didn't know t is mistake.

drspartez: Ok

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/JqlMemoryTest.kt
wyrzyk: I think runtime and dependency costs doesn't need explanation. I want to add something about readability and diagnosability:

- A proper mock could be reused across multiple tests. We have just two tests here, but we need to repeat the blocks of code that creates the mock.
- It can introduce bugs in the future. Without Mockito, when changing mocked objects, you need to adjust your mocks. You need to make a conscious decision, how the mock should behave.  Mockito can transparently adapt for most cases. In the end, you may end up testing something else than intended (we hade examples of the cases like that internally). 
- Like every reflection-based tool, it may fool IDE. Searching for usages of a method may not work well (I'm not sure if it's still an issue in IJ 2019, but it's an open-source project and developers can use any IDE).
- If you put complex logic into your mock, then it's hard to debug (reflection again). It's ok if you return values, but by introducing the library on a classpath, you encourage developers to use it for every case.
- We have multiple reasons to write tests. To protect against regressions. To document how to use an API. To reproduce the issue. Another reason is to verify if the API is useable. Mockito makes mocking unmockable objects easy. It means we'll stop to see problematic places and we'll never improve it. Problems with testing are very similar to issues of the code and reuse of it. We should treat our tests as the first user of an API.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/memories/JqlMemory.kt
```diff
@@ -1,7 +1,14 @@
...
 import com.atlassian.performance.tools.jiraactions.api.page.IssuePage
+import java.util.function.Predicate
 
 interface JqlMemory : Memory<String> {
+
     fun observe(issuePage: IssuePage)
+
+    @JvmDefault
+    fun recallByTag(filter: Predicate<String>): String? {
+        return null;
```
lipinskipawel: the semicolon at the of the `return null`

wyrzyk: I'm also concerned about returning `null` as a default implementation. It will make the change source compatible, but I see no value besides that. It'll probably hide the problem until it appears in a different place (no results for the action at the end of the test).

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/memories/JqlMemory.kt
wyrzyk: @drspartez When making a tradeoff like this, it's a good practice to describe it in a commit's description. It adds a couple of extra benefits:
- You need to document the decision, so you need to make a logically sound block of text. For some persons (including me) it helps to sort things out. Sometimes it may affect a decision.
- Other people can review your decision. It'll help them to understand it and may help to find wholes in the reasoning. It may also affect the decision.
- If a change is confusing for the current readers, then it'll be confusing for future readers too. Commits are easier to access than PR.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/memories/JqlMemory.kt
lipinskipawel: Maybe this method should be documented e.g which strings are allowed? Can we somehow validate them? Should we change the name of the method into recallByPredicate? We can throw an Exception as the default implementation.

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/JqlMemoryTest.kt
drspartez: > * A proper mock could be reused across multiple tests. We have just two tests here, but we need to repeat the blocks of code that creates the mock.

I used to create mock to cover some specific code path and i think it is right.
You're right about redundant mock preparation code.

>* It can introduce bugs in the future. Without Mockito, when changing mocked objects, you need to adjust your mocks. You need to make a conscious decision, how the mock should behave. Mockito can transparently adapt for most cases. In the end, you may end up testing something else than intended (we had examples of the cases like that internally).

Mockito requires using explicit API calls to prepare mocked instance so it is cean how API is used and mocked. Involving mockito or not has nothing to do with test scope/goal/intention.

> * Like every reflection-based tool, it may fool IDE. Searching for usages of a method may not work well (I'm not sure if it's still an issue in IJ 2019, but it's an open-source project and developers can use any IDE).

There are explicit API calls

> * If you put complex logic into your mock, then it's hard to debug (reflection again). It's ok if you return values, but by introducing the library on a classpath, you encourage developers to use it for every case.

There is no complex logic in here

> * We have multiple reasons to write tests. To protect against regressions. To document how to use an API. To reproduce the issue. Another reason is to verify if the API is useable. Mockito makes mocking unmockable objects easy. It means we'll stop to see problematic places and we'll never improve it. Problems with testing are very similar to issues of the code and reuse of it. We should treat our tests as the first user of an API.

This is separate concern (unmockable objects and its improvement) and mockito does not caused the problem and does not solves it (in terms of design). When you wrote dozen lines of code without mockito and it is working for you, then this also causes problematic places are not visible.
Mockito helps to show how api can be used . Ido not see this is broken in area of test concern

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/memories/JqlMemory.kt
drspartez: Default recall also returns null when there is no result i think it is better than exception.
In my opinion name is ok because it provides more context. Everybody see that predicate is there

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/memories/JqlMemory.kt
drspartez: This is something new to me in scope of commit message. I was used to deal with project/design decisions on issue level. But here it is brainstormed more on PR. So i don't get why commit message appears here.

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/JqlMemoryTest.kt
wyrzyk: >I used to create mock to cover some specific code path and i think it is right.
You're right about redundant mock preparation code.

It's ok. It just violates the DRY principle. For some cases, redundancy can be better than a broken abstraction. I would not put much weight to that.

> Mockito requires using explicit API calls to prepare mocked instance so it is cean how API is used and mocked. Involving mockito or not has nothing to do with test scope/goal/intention.

This is an internal example of the problem I've mentioned https://hello.atlassian.net/wiki/spaces/~kcichy/blog/2018/10/12/321652230/Do+you+trust+your+mock#Mocks-go-wrong-over-time. It's clear how to use and mock objects. It'll work for most cases at the time when the test is written. But Mockito helps mocks to deteriorate unnoticeably.


> There is no complex logic in here


I'm not concerned with complex logic here. I'm concerned about giving a tool (on a classpath) that can be misused in the future.

> This is separate concern (unmockable objects and its improvement) and mockito does not caused the problem and does not solves it (in terms of design). When you wrote dozen lines of code without mockito and it is working for you, then this also causes problematic places are not visible.
Mockito helps to show how api can be used . Ido not see this is broken in area of test concern

 
Mockito makes dealing with problematic API easy. Writing dozens of lines of code is painfull, so it reminds about the problem. It's still easier to write the mocks than to fix the API, but the problem lays in a different place. The problem may be invisible (it's easy to mock with Mockito) for a new code. We may introduce a new broken API that's hard to mock because we have a tool with super-powers.

drspartez: @wyrzyk 
I do not see that reason for the problem with test mentioned in the blogpost of Kamil was mockito. If the test was written with manual mocks results would be the same.
Primary reason that the test didn't catch the problem is not implemented coverage.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/memories/JqlMemory.kt
wyrzyk: > So i don't get why commit message appears here.

It appears here to indicate that we could avoid confusion like in this thread in the future if we document decisions. I also added reasons why it's a good practice to document it close to the source. 


decision documented in a commit description >> decision documented in the issue >> decision documented in the PR >> decision not documented at all.

For some cases, code comments would be best for documenting a decision. I'd leave it out of the scope of the discussion, as it highly depends on the context.

drspartez: @wyrzyk Would You like to propose neat pattern for such description relevant for commit message ?

wyrzyk: I'm not the best person in case of writing human-readable text, but I'll try:
```
JPERF-522: Add support for filtering by JQL type in JqlMemory

JqlMemory operates on Strings instead of JQL objects. It makes evolving the API
hard. I decided to keep using String for this change to avoid API refactoring costs.
```

drspartez: Cool. Thanks

mcmurphy2000:
Guys we need to move with this, I think there was enough discussion, some things were resolved, there some disagreement but it's blocking our current sprint. We need to move on.

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/JqlMemoryTest.kt
drspartez: All test stuff is reverted. Not because I do agree with this review, but because I'm forced by time.

drspartez:
Test to this code will appear with next PR related to JQL action split.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/memories/adaptive/AdaptiveJqlMemory.kt
```diff
@@ -5,48 +5,61 @@ import com.atlassian.performance.tools.jiraactions.api.memories.JqlMemory
...
     }
+
+    override fun recallByTag(filter: Predicate<String>): String? {
+        return random.pick(jqls.filter {it.tag != null && filter.test(it.tag.toString()) }.toList())?.baked?.jql
+    }
+
+    private class TaggedBakedJql(val baked: BakedJql, val tag: String? = null)
 }
 
 data class BakedJql(
```
lipinskipawel: Since we are doing a minor change, perhaps we can deprecate this class? 
Why this class should stay public?

drspartez: I see this class as candidate to deprecate when we introduce some String with JQL representation class - it would be exact replacement.

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/JqlMemoryTest.kt
dagguh: @drspartez I think there's been a misunderstanding. I agreed that Mockito is necessary in this case and that we should go for it. Having a test with Mockito is better than not having a test at all.

I meant to inform that we wouldn't have to use Mockito if `IssuePage` was better designed. We can improve its design only in a major release. Then we should remove Mockito.

Being forced to resort to reflection hacks like Mockito proves how hard it is to customize `IssuePage`. Currently users have to settle on copy-pasting implementations.

TL;DR: Mockito is bad, but this time it was necessary.

PR #18, opened 2019-07-16
JPERF-525: Add new action search Jql "a*"

lipinskipawel:
https://ecosystem.atlassian.net/browse/JPERF-525


src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/JiraCoreScenario.kt
```diff
@@ -76,6 +82,7 @@ class JiraCoreScenario constructor() : Scenario {
         val actionProportions = mapOf(
             createIssue to 5,
             searchWithJql to 20,
+            searchWithJqlStar to 6,
```
mcmurphy2000: I'm afraid this can be considered as behavioral change and therefore API breaking :frowning:

lipinskipawel: Sure, I have just added this in changelog

sebapawlak: I agree. Let's not change the behaviour of the default scenario. We can change the behaviour in our custom scenario.

sebapawlak:
Changelong entry missing

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/action/SearchJqlStarAction.kt
```diff
@@ -0,0 +1,37 @@
...
+    private val jira: WebJira,
+    private val meter: ActionMeter,
+    private val jqlMemory: JqlMemory,
+    private val issueKeyMemory: IssueKeyMemory
+) : Action {
+    override fun run() {
+        val jqlQuery = jqlMemory.recallByTag(Predicate { s -> s == "GENERIC_WIDE" })!!
+
+        val issueNavigatorPage = meter.measure(
+            key = SEARCH_WITH_JQL,
```
mcmurphy2000: I wonder if we should have a separate key for this action. Otherwise, how will we distinguish it from other "Search with JQL" actions later?

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/action/SearchJqlStarAction.kt
```diff
@@ -0,0 +1,37 @@
...
+import javax.json.JsonObject
+
+class SearchJqlStarAction(
+    private val jira: WebJira,
+    private val meter: ActionMeter,
+    private val jqlMemory: JqlMemory,
+    private val issueKeyMemory: IssueKeyMemory
+) : Action {
+    override fun run() {
+        val jqlQuery = jqlMemory.recallByTag(Predicate { s -> s == "GENERIC_WIDE" })!!
```
mcmurphy2000: I think we need to account for `null` here. What if there are no such queries?

lipinskipawel: This query is [hardcoded](https://github.com/atlassian/jira-actions/blob/master/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/memories/adaptive/AdaptiveJqlMemory.kt#L25).

dagguh:
Nice!

PR #20, opened 2019-07-17
Add a scenario which aggregates all actions at a given time

lipinskipawel:
The purpose of this PR is to have a discussion. I'm not going to merge this as this is even a draft. Any feedback and ideas are welcome as usual.

### DIFF
--- a/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/JiraAllActionScenario.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/JiraAllActionScenario.kt
@@ -1,0 +1,130 @@
+package com.atlassian.performance.tools.jiraactions.api.scenario
+
+import com.atlassian.performance.tools.jiraactions.api.SeededRandom
+import com.atlassian.performance.tools.jiraactions.api.WebJira
+import com.atlassian.performance.tools.jiraactions.api.action.*
+import com.atlassian.performance.tools.jiraactions.api.measure.ActionMeter
+import com.atlassian.performance.tools.jiraactions.api.memories.IssueKeyMemory
+import com.atlassian.performance.tools.jiraactions.api.memories.adaptive.AdaptiveIssueKeyMemory
+import com.atlassian.performance.tools.jiraactions.api.memories.adaptive.AdaptiveIssueMemory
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/JiraAllActionScenario.kt
```diff
@@ -0,0 +1,130 @@
...
+                issueKeyMemory = AdaptiveIssueKeyMemory(seededRandom)
+            }
+        }
+    }
+
+    /**
+     * You can use `Builder` to share memories between Scenarios.
+     * @since 3.6.0
+     */
+    class Builder {
```
drspartez: This is second place with builder of the same logical purpose. We may consider extraction and refactoring. I mean builder as parametererized type (by proper Scenario implementation class)

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/JiraAllActionScenario.kt
```diff
@@ -0,0 +1,130 @@
...
+
+/**
+ * Provides Jira `Scenario`.
+ * This class returns all available actions at current time.
+ * It is highly possible and expected that over time these actions
+ * will be different. Thus <strong>don't</strong> assume
+ * actions provided by this class will never change.
+ * @since 3.6.0
+ */
+class JiraAllActionScenario constructor() : Scenario {
```
mcmurphy2000: Am I correct in my understanding that the only reason why we're creating this class is because we don't have access to `issueKeyMemory` outside of this module?
 If so, can't we just create a custom scenario similar to the way we're creating [JiraSoftwareScenario](https://github.com/atlassian/jira-software-actions/blob/b5e9bda7845bdeea4dc52568de52714b8ca1cb3c/src/main/kotlin/com/atlassian/performance/tools/jirasoftwareactions/api/JiraSoftwareScenario.kt#L31) ?

drspartez: We have acceess to issueKeyMemory. But the scenario itself has own memory, and the constructor is used to provide externally initialized (then the built in one is not used),

PR #23, opened 2019-07-25
JQL refactoring draft

drspartez:
In the meantime i baked JQL related refactiring proposal.

I've divided JQL domain into 3 public interfaces:
* Jql - instance of prepared query 
* JqlSuplier - denotes a 'brain' that knows how to prepare query (former 'PrescriptionType')
* JqlContext -  runtime data to prepare query

There are also public **JqlMemory2**, **AdaptiveJqlMemory2** and  (names just to distinguish from previous)
Other related stuff are implementation


src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/jql/Jql.kt
```diff
@@ -0,0 +1,19 @@
+package com.atlassian.performance.tools.jiraactions.api.jql
+
+import com.atlassian.performance.tools.jiraactions.api.SeededRandom
+import com.atlassian.performance.tools.jiraactions.api.page.IssuePage
+
+interface JqlContext {
```
sebapawlak: Looks like and public SPI. Historically we have many issues with SPI written in Kotlin and how it interoperated with Java and Java interfaces and their default methods. I would suggest writhing this part in Java which should not be costly but may pay-off in case we want to extend the SPI in the future without breaking the compatibility.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/memories/adaptive/AdaptiveJqlMemory2.kt
```diff
@@ -0,0 +1,68 @@
...
+import com.atlassian.performance.tools.jiraactions.api.memories.adaptive.jql.JqlPrescription
+import com.atlassian.performance.tools.jiraactions.api.memories.adaptive.jql.JqlPrescriptions
+import com.atlassian.performance.tools.jiraactions.api.page.IssuePage
+import com.atlassian.performance.tools.jiraactions.jql.BuiltInJQL
+import com.atlassian.performance.tools.jiraactions.jql.JqlContextImpl
+import org.apache.logging.log4j.LogManager
+import org.apache.logging.log4j.Logger
+import java.util.function.Predicate
+
+class AdaptiveJqlMemory2(
```
lipinskipawel: Why this class is public? We have already public interface JqlMemory2.
Should we create some sort of factory that returns the instance of JqlMemory2 in this case AdaptiveJqlMemory2?
What should be a general idea here?

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/jql/Jql.kt
```diff
@@ -0,0 +1,19 @@
...
+    fun seededRandom(): SeededRandom
+    fun issuePage(): IssuePage?
+}
+
+interface JqlSupplier {
```
lipinskipawel: Is it possible to reduce the number of public interfaces? It seems that our API grows quickly.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/memories/adaptive/AdaptiveJqlMemory2.kt
drspartez: in general classes with "2" suffix are replacements, for existing which will be deprecated and removed with major change, so in some time number of interfaces will be the same as before refactoring

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/jql/Jql.kt
drspartez: supplier is replacement for PrescriptionType which is public and coukld be removed only witch major change

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/memories/adaptive/AdaptiveJqlMemory2.kt
drspartez: @lipinskipawel for more invasive refactoring You can look here
https://github.com/drspartez/jira-actions/tree/issue/memories-refactoring

lipinskipawel: My question was more about the Factory class, I do understand the "2" suffix.
Anyway, good to see Factory in the commit you sent.

drspartez: OK. So concern of this change is to replace "String" jql representation with something else which is more relevant (so all JQL interfaces ore for this reason). Other stuff remains almost untouched (so we still offer Adaptive memories implementations as usual). 
In second approach (https://github.com/drspartez/jira-actions/tree/issue/memories-refactoring) i changed the fashion we provide adaptive memories (hidden implementation and factories), also reduced number of actual memory interfaces to 2.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/memories/adaptive/AdaptiveJqlMemory2.kt
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    private val preparedQueries = mutableListOf<Jql>();
+
+    init {
+        BuiltInJQL.RESOLVED.get(JqlContextImpl(random))?.let { preparedQueries.add(it) }
+        BuiltInJQL.GENERIC_WIDE.get(JqlContextImpl(random))?.let { preparedQueries.add(it) }
+    }
+
+    private val availableJqlSuppliers = mutableSetOf(
```
dagguh: `Memory` impls are implicitly allowed to be not thread-safe, because each VU has its own memories.
It's a nice opportunity to make that assumption explicit and mark `AdaptiveJqlMemory2` as `@NotThreadSafe`.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/memories/adaptive/AdaptiveJqlMemory2.kt
dagguh: > Why this class is public?

The old `AdaptiveJqlMemory` was also API, so it's perfectly fine for the replacement to be public as well.
It was published to make custom scenarios easier to write.

dagguh: > Should we create some sort of factory that returns the instance of JqlMemory2 in this case AdaptiveJqlMemory2?

Factories cost complexity and must have additional value to be justified. AFAIK across all JPT libs we expose only one factory: `com.atlassian.performance.tools.report.api.result.RawCohortResult.Factory`. We did that to offer different kinds of `RawCohortResult` without allowing custom subtypes.

Why did we forbid custom subtypes? Because we needed to add methods to the (old) `CohortResult`, but we didn't have a pattern for that. So we created a replacement (just like `JqlMemory2`) and we wanted to protect ourselves from this problem in the future. We decided nobody should need to implement this interface. In short, we wanted `RawCohortResult` to be API, but not SPI. 

The question is - do we want to allow custom implementations of `JqlMemory2`? If so, let's go for a factory, else public constructors/builders.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/memories/adaptive/AdaptiveJqlMemory2.kt
```diff
@@ -0,0 +1,68 @@
...
+        BuiltInJQL.GIVEN_WORD
+    )
+
+    override fun observe(issuePage: IssuePage) {
+        val jql = availableJqlSuppliers.asSequence()
+            .map { it.get(JqlContextImpl(random, issuePage)) }
+            .filter { it != null }
+            .firstOrNull()
+
+        jql?.let {
```
dagguh: Micro-code-style 😉
`jql` can be inlined and therefore chained
and then the `it` can be renamed to `jql`

src/main/kotlin/com/atlassian/performance/tools/jiraactions/jql/JqlImpl.kt
```diff
@@ -0,0 +1,87 @@
...
+import com.atlassian.performance.tools.jiraactions.api.jql.JqlSupplier
+import com.atlassian.performance.tools.jiraactions.api.page.IssuePage
+
+internal class JqlImpl(private val query: String,
+                       private val supplier: JqlSupplier): Jql {
+    override fun query(): String = this.query
+    override fun supplier(): JqlSupplier = this.supplier
+}
+
+internal class JqlContextImpl(private val rnd: SeededRandom, private val page: IssuePage?) : JqlContext {
```
dagguh: `rnd` - Research and Development? :trollface:

src/main/kotlin/com/atlassian/performance/tools/jiraactions/jql/JqlImpl.kt
```diff
@@ -0,0 +1,87 @@
...
+    PROJECT_ASSIGNEE({
+        ctx -> ctx.issuePage()?.let {
+        val userName = it.getAssignee()
+        val project =  it.getProject()
+        if (userName != null && project != null) {
+            "project = ${project.key} and assignee = $userName order by reporter"
+        } else {
+            null
+        }
+    }
```
dagguh: There's something wonky about the indents here

src/main/kotlin/com/atlassian/performance/tools/jiraactions/jql/JqlImpl.kt
```diff
@@ -0,0 +1,87 @@
+package com.atlassian.performance.tools.jiraactions.jql
+
+import com.atlassian.performance.tools.jiraactions.api.SeededRandom
+import com.atlassian.performance.tools.jiraactions.api.jql.Jql
+import com.atlassian.performance.tools.jiraactions.api.jql.JqlContext
+import com.atlassian.performance.tools.jiraactions.api.jql.JqlSupplier
+import com.atlassian.performance.tools.jiraactions.api.page.IssuePage
+
+internal class JqlImpl(private val query: String,
```
dagguh: "JQL implementation". What makes this one different from the other implementations?
If none other exist, how *could* they differ?

It seems that the interface allows dynamic `query` and `supplier` logic, but this impl simply reuses the provided fields. Suggestions: `StaticJql`, `SuppliedJql`.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/memories/JqlMemory2.kt
```diff
@@ -0,0 +1,12 @@
+package com.atlassian.performance.tools.jiraactions.api.memories
+
+import com.atlassian.performance.tools.jiraactions.api.jql.Jql
+import com.atlassian.performance.tools.jiraactions.api.page.IssuePage
+import java.util.function.Predicate
+
+interface JqlMemory2 : Memory<Jql> {
```
dagguh: If this a replacement for `JqlMemory`, shouldn't that one be deprecated?
If it's too early to deprecate, then maybe it's too early to put in replacements? New API works best if it was tried in some way, e.g.:
* unit/integration tests within the module
* equivalent API used outside of the module

PR #26, opened 2019-08-09
Issue/jperf 344 disable backup service

lipinskipawel:


CHANGELOG.md
```diff
@@ -26,6 +26,11 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/jira-actions/compare/release-3.7.0...master
 
+### Fixed
+- Disable backup service during the test [JPERF-344].
```
wyrzyk: Please keep the changelog conventions. 
`Blah blah. Resolve [JPERF-XXX].`
😄

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/BackupConfiguration.kt
```diff
@@ -0,0 +1,46 @@
+package com.atlassian.performance.tools.jiraactions.api.page
+
+import com.atlassian.performance.tools.jiraactions.Patience
+import org.apache.logging.log4j.LogManager
+import org.openqa.selenium.By
+import org.openqa.selenium.WebDriver
+
+class BackupConfiguration(
```
wyrzyk: Do we need/want to expose it in API?

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/BackupConfiguration.kt
```diff
@@ -0,0 +1,46 @@
...
+    private val driver: WebDriver,
+    private val access: AdminAccess
+) {
+    private val logger = LogManager.getLogger(this::class.java)
+    private val backupLocator = By.id("del_10001")
```
wyrzyk: Looks like a generated and data specific id. Is it always the same on all Jira instances over the world?

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/BackupConfiguration.kt
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     * Makes sure backup service is disabled
```
wyrzyk: Shouldn't it be named `makeSureIsDisabled` then? Now Javadoc suggests one thing and the method name another.

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/BackupConfigurationIT.kt
```diff
@@ -0,0 +1,86 @@
...
+import org.junit.Test
+import org.openqa.selenium.By
+import java.time.Clock
+import java.util.*
+
+class BackupConfigurationIT {
+    private val logger: Logger = LogManager.getLogger(this::class.java)
+
+    /**
+     * During the test, you can connect to WebDriver by the VNC viewer.
```
wyrzyk: ❤️

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/BackupConfiguration.kt
lipinskipawel: it should be internal

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/BackupConfiguration.kt
dagguh: It does seem weirdly specific. If we can find a more stable-looking selectors, let's use that.
Otherwise, let's use this one. If we detect bugs with this, we'll react and fix, using the bug details as guidance.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/BackupConfiguration.kt
lipinskipawel: This JavaDoc is redundant, I will remove it. The method name is clear.

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/BackupConfigurationIT.kt
```diff
@@ -0,0 +1,86 @@
...
+     * DockerisedChrome opens the port 5900 so executing command
+     * docker ps
+     * will show you which port on your host is open and mapped to 5900 on the container.
+     */
+    @Test
+    fun backupShouldBeDisable() {
+        val version = System.getenv("JIRA_SOFTWARE_VERSION") ?: "8.0.0"
```
dagguh: Awesome! Not only we have a test, but also it's covered by multiple Jira versions.

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/BackupConfigurationIT.kt
```diff
@@ -0,0 +1,86 @@
...
+    private val logger: Logger = LogManager.getLogger(this::class.java)
+
+    /**
+     * During the test, you can connect to WebDriver by the VNC viewer.
+     * DockerisedChrome opens the port 5900 so executing command
+     * docker ps
+     * will show you which port on your host is open and mapped to 5900 on the container.
+     */
+    @Test
+    fun backupShouldBeDisable() {
```
wyrzyk: I'd recommend testing the scenario on one of our internal datasets (use snapshot version of jira-actions). We would have it tested on multiple configurations.

dagguh: Naming tip: `BackupConfigurationIT.backupShouldBeDisable` reads:
> Backup configuration backup should be disable

If we rename it to `BackupConfigurationIT.shouldDisableBackups`, it'd read:
> Backup configuration should disable backups

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/BackupConfiguration.kt
dagguh: It's a copy-paste from `RichTextEditorConfiguration`:
```
/**
 * Makes sure RTE is disabled, even if the feature doesn't exist at all.
 * Supports both enabled and disabled websudo.
 */
```
`ensureRteIsDisabledRegardlessOfJiraVersionOrWebsudo` would tell the whole story, but it's not worth reminding that to every reader, caller or stacktrace. It's ok to put fine details in the docs.

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/BackupConfigurationIT.kt
lipinskipawel: Done

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/BackupConfiguration.kt
```diff
@@ -0,0 +1,49 @@
...
+            "//*[contains(text(), 'Delete')]"
+    )
+
+    fun disable(): BackupConfiguration {
+        val prompted = access.isPrompted()
+        if (prompted) {
+            access.gain()
+        }
+        if (isBackupServicePresent()) {
+            ensureSwitchIsOff()
```
dagguh: There's no switch on the page. This is copy-pasted from `RichTextEditorConfiguration`, which has a switch: 
image

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/BackupConfiguration.kt
```diff
@@ -0,0 +1,49 @@
+package com.atlassian.performance.tools.jiraactions.api.page
```
wyrzyk: If it's not API we can keep it outside of the `api` package.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/BackupConfiguration.kt
```diff
@@ -0,0 +1,49 @@
...
+            "//*[contains(@class, 'operations-list')]" +
+            "//*[contains(text(), 'Delete')]"
+    )
+
+    fun disable(): BackupConfiguration {
+        val prompted = access.isPrompted()
+        if (prompted) {
+            access.gain()
+        }
+        if (isBackupServicePresent()) {
```
dagguh: No need for the `if`.
If `driver.findElements(backupLocator)` is empty, then `.forEach { it.click() }` will work just fine.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/BackupConfiguration.kt
```diff
@@ -0,0 +1,49 @@
...
+        return this
+    }
+
+    private fun isBackupServicePresent(): Boolean {
+        return Patience().test { driver.isElementPresent(backupLocator) }
+    }
+
+    private fun ensureSwitchIsOff() {
+        driver.findElements(backupLocator).forEach { it.click() }
+        driver.switchTo().alert().accept()
```
dagguh: Won't this pop up after each `it.click()`?

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/BackupConfiguration.kt
```diff
@@ -0,0 +1,49 @@
...
+        }
+        return this
+    }
+
+    private fun isBackupServicePresent(): Boolean {
+        return Patience().test { driver.isElementPresent(backupLocator) }
+    }
+
+    private fun ensureSwitchIsOff() {
+        driver.findElements(backupLocator).forEach { it.click() }
```
dagguh: The integration test should prepare multiple backup services, because it seems we're testing with only one.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/BackupConfiguration.kt
```diff
@@ -0,0 +1,49 @@
...
+        driver.switchTo().alert().accept()
+        if (access.isPrompted()) {
```
dagguh: The admin access pops up after we delete a backup service?

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/BackupConfiguration.kt
```diff
@@ -0,0 +1,49 @@
...
+import org.apache.logging.log4j.LogManager
+import org.openqa.selenium.By
+import org.openqa.selenium.WebDriver
+
+internal class BackupConfiguration(
+    private val driver: WebDriver,
+    private val access: AdminAccess
+) {
+    private val logger = LogManager.getLogger(this::class.java)
+    private val backupLocator = By.xpath(
```
dagguh: Actually it's a `deleteBackupLocator`

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/BackupConfiguration.kt
```diff
@@ -0,0 +1,49 @@
+package com.atlassian.performance.tools.jiraactions.api.page
+
+import com.atlassian.performance.tools.jiraactions.Patience
+import org.apache.logging.log4j.LogManager
+import org.openqa.selenium.By
+import org.openqa.selenium.WebDriver
+
+internal class BackupConfiguration(
```
dagguh: If it's `internal`, we should move it outside `api` subpackage, to `com.atlassian.performance.tools.jiraactions.page`

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/BackupConfiguration.kt
lipinskipawel: Yes, I have seen this view VNC.

dagguh: Does it pop after every deletion? Current code handles it *after* clicking *all* deletions.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/action/SetUpAction.kt
```diff
@@ -11,6 +11,7 @@ class SetUpAction(
     override fun run() {
         meter.measure(SET_UP) {
             jira.configureRichTextEditor().disable()
+            jira.configureBackupPolicy().disable()
```
dagguh: This is quite a destructive operation. It actually `delete()`s backup services.
This `SetUpAction` is used by default in most APIs, including app vendors and customers. I believe we shouldn't destroy Jira information by default. It should be possible to opt-in to, e.g. by exposing this new API and letting users customize their scenarios.

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/BackupConfigurationIT.kt
```diff
@@ -0,0 +1,86 @@
...
+    private val logger: Logger = LogManager.getLogger(this::class.java)
+
+    /**
+     * During the test, you can connect to WebDriver by the VNC viewer.
+     * DockerisedChrome opens the port 5900 so executing command
+     * docker ps
+     * will show you which port on your host is open and mapped to 5900 on the container.
+     */
+    @Test
+    fun shouldDisableBackups() {
```
dagguh: The CI fails with "Gradle exited with 137", which means the OS is running out of memory.

I'll run `./gradlew build` locally and see how much memory is consumed and how long it takes.
The memory can be used by Gradle and Docker.
If it's longer than 21 minutes, we'll need to bump the `travis_wait` in `.travis.yml`.

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/BackupConfigurationIT.kt
dagguh: This is so good!
PS. the [VNC password is `secret`](https://github.com/testcontainers/testcontainers-java/blob/06ea95be1e288622aedb3a35ea585575654e5a9a/core/src/main/java/org/testcontainers/containers/VncRecordingContainer.java#L33). Open-source FTW :)

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/BackupConfigurationIT.kt
dagguh: Ok, the local run took: `BUILD SUCCESSFUL in 19m 35s`
Before the test, Docker used 1 GB. During the test it used 6 GB. The containers were nicely cleaned up after every test, but Docker did not release the memory. It's been 5 minutes after the tests have finished and it still holds 6 GB.

dagguh: PS. I observed the build via VNC, we're accidentally testing it multiple times. `JiraCoreScenarioIT` runs the `scenario.getSetupAction` already. Perhaps integration tests are truly about integration - joining things together. In this case we could hijack the test and cover the backup service deletion, even if it's not default anymore.

dagguh: In the VNC, I also saw this:
image
image
As if a previous `sendKeys(license)` kept a buffers of keys to send and we navigated away and kept sending it on the next screen.

dagguh: WRT memory, it seems we can tell [Java to work better with virtualization](https://stackoverflow.com/a/51364393). Note that this was about running Java in Docker, and we're running Docker via Java. *But* the CI environment is also virtualized, so it might help.
Additionally, we can set the max heap for the JUnit JVM. It seems our Travis has 7.5 GB of RAM. If 5-6 GB go for Docker, then we're left with around 1 GB for the JVM. So let's [set the max heap](https://github.com/atlassian/jira-hardware-exploration/blob/6efaff838cf59fad1e4c77276e90051827e862d0/build.gradle.kts#L41) to 1 GB or less.

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/BackupConfigurationIT.kt
```diff
@@ -0,0 +1,86 @@
...
+class BackupConfigurationIT {
+    private val logger: Logger = LogManager.getLogger(this::class.java)
+
+    /**
+     * During the test, you can connect to WebDriver by the VNC viewer.
+     * DockerisedChrome opens the port 5900 so executing command
+     * docker ps
+     * will show you which port on your host is open and mapped to 5900 on the container.
+     */
+    @Test
```
dagguh: Starting another convo for the test duration.
Here's my local report: 
image

We should either start merging the integration tests or bump the `travis_wait`

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/BackupConfigurationIT.kt
lipinskipawel: I’m using `docker stats` and according to that tool my containers are not using more memory than 2.5 GB (first ~2GB, second ~400 MB) per one test.
Docker version 19.03.1

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/BackupConfigurationIT.kt
lipinskipawel: I will bump the `travis_wait` maybe we should bump to 7 minutes to avoid flakiness?
or there is no such danger?

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/BackupConfiguration.kt
lipinskipawel: This class will be part of the API.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/BackupConfiguration.kt
lipinskipawel: This class will be part of the API.

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/BackupConfigurationIT.kt
dagguh: 7 minutes sound good

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/TestingScenario.kt
```diff
@@ -0,0 +1,76 @@
...
+import com.atlassian.performance.tools.jiraactions.api.action.*
+import com.atlassian.performance.tools.jiraactions.api.measure.ActionMeter
+import com.atlassian.performance.tools.jiraactions.api.memories.adaptive.AdaptiveIssueKeyMemory
+import com.atlassian.performance.tools.jiraactions.api.memories.adaptive.AdaptiveIssueMemory
+import com.atlassian.performance.tools.jiraactions.api.memories.adaptive.AdaptiveJqlMemory
+import com.atlassian.performance.tools.jiraactions.api.memories.adaptive.AdaptiveProjectMemory
+
+class TestingScenario : Scenario {
+
+    override fun getActions(jira: WebJira, seededRandom: SeededRandom, meter: ActionMeter): List<Action> {
```
dagguh: `TestingScenario.getActions` isn't called.
Let's remove it and simply run `jira.configureBackupPolicy().disable()`.

Previously I was considering merging ITs together, which would cover "backups vs actions" case for free. But we decided to split integration tests per feature, so the case is no longer worth it.

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/BackupConfigurationIT.kt
```diff
@@ -0,0 +1,168 @@
...
+        val actionMeter = ActionMeter(
+            virtualUser = UUID.randomUUID(),
+            output = CollectionActionMetricOutput(metrics),
+            clock = Clock.systemUTC(),
+            w3cPerformanceTimeline = DisabledW3cPerformanceTimeline()
+        )
+        val user = User("admin", "admin")
+        val userMemory = object : UserMemory {
+            override fun recall(): User {
+                return user
```
dagguh: You can use the `Convert to expression body` IDE action on both `UserMemory` functions to the impl shorter.
`shouldDisableBackups` is very long and methods should fit on one screen entirely.

CHANGELOG.md
```diff
@@ -26,6 +26,11 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/jira-actions/compare/release-3.8.0...master
 
+### Fixed
+- Disable backup service during the test. Resolve [JPERF-344].
```
dagguh: What's "the test"?

The diff shows that we're *adding* `WebJira.configureBackupPolicy` and `BackupConfiguration.disable`.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/BackupConfiguration.kt
```diff
@@ -0,0 +1,76 @@
...
+    )
+
+    /**
+     * By default, this function will try to delete 10 backup services.
+     * If your instance has less than 10 backup services it will remove all of them.
+     * If your instance has more than 10 backup services it will remove only 10.
+     * Recommended function to use in case of more than 10 backups is {@link #disable(Int)}
+     * where you can explicitly set a number of services to delete.
+     */
+    fun disable(): BackupConfiguration {
```
dagguh: Let's call it `delete`.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/BackupConfiguration.kt
```diff
@@ -0,0 +1,76 @@
...
+    private val deleteBackupLocator = By.xpath(
+        "//*[@id='tbl_services']" +
+            "//*[contains(text(), 'com.atlassian.jira.service.services.export.ExportService')]" +
+            "/ancestor::tr" +
+            "//*[contains(@class, 'operations-list')]" +
+            "//*[contains(text(), 'Delete')]"
+    )
+
+    /**
+     * By default, this function will try to delete 10 backup services.
```
dagguh: Where did the "limit 10" come from?
We're deleting backup services for a reason, why would it be ok to leave *some* behind?

build.gradle.kts
```diff
@@ -77,6 +77,7 @@ tasks.getByName("test", Test::class).apply {
 
 val testIntegration = task<Test>("testIntegration") {
     include("**/*IT.class")
+    maxHeapSize = "8g"
```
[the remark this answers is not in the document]
lipinskipawel: There is 1g already. I have changed it.

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/TestingScenario.kt
lipinskipawel: I can not remove `getActions` because it is mandatory to override this function.
Do you propose to return null? What is the point in that case even if this function isn't called anywhere?

dagguh: I propose to remove `TestingScenario`

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/BackupConfiguration.kt
lipinskipawel: Here, I followed the pattern in `RichTextEditorConfiguration`, but in this case, the actual action is just `delete`. I will rename it.

dagguh: Well, the RTEC actually does *disable* RTE and it doesn't *delete* RTE. It's reversible, not destructive.

CHANGELOG.md
lipinskipawel: Due to not enabled this feature by default I will change the message to 
`### Added
- Add ability to delete service backup. Resolve [JPERF-334].`

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/BackupConfiguration.kt
dagguh: PS. RTEC describes an actual page - there is such a thing as "Rich Text Editor configuration" page. There is no "backup configuration" page. But we can leave that in and pretend the "Services" page acts like a backup config page. We can always add the actual services page in the future.

CHANGELOG.md
dagguh: Sounds good!

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/TestingScenario.kt
lipinskipawel: Ok, so I will remove it and modify `JiraEditScenario` and add a similar class to `TestingSetup` to the `RichTextEditorIT` file.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/BackupConfiguration.kt
lipinskipawel: Leaving some backup services is not ok. That is why there is another function in the API.
Why do we even need such a limit?
To avoid potentials problems with never-ending loop.

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/TestingScenario.kt
dagguh: `RichTextEditorIT` tests edit actions, because they're impacted by `RichTextEditorTextArea` and `RichTextEditorConfiguration` - they use the RTE.
What do we need `JiraEditScenario` for here?

lipinskipawel: Aha, I assumed wrongly that my test needs `Scenario` which is no true. I can fire `TestingSetup` action manually.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/BackupConfiguration.kt
dagguh: If it's not ok, then this method is not ok and we need to remove it.
If infinite loops are real then the other method is not ok and we need to remove that.
Let's be consistent here.

Also, suppose the infinite loop is real - then it means the deletions don't work. So we give up after 10 attempts and carry on? The user asked to delete backup services and we didn't do it and silently ignored that failure.
Let's just remove this weird complication. If infinite loops will happen, then timeouts will trip up.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/BackupConfiguration.kt
```diff
@@ -0,0 +1,42 @@
...
+    private val logger = LogManager.getLogger(this::class.java)
+    private val deleteBackupLocator = By.xpath(
+        "//*[@id='tbl_services']" +
+            "//*[contains(text(), 'com.atlassian.jira.service.services.export.ExportService')]" +
+            "/ancestor::tr" +
+            "//*[contains(@class, 'operations-list')]" +
+            "//*[contains(text(), 'Delete')]"
+    )
+
+    fun delete(): BackupConfiguration {
```
dagguh: To differentiate from `deleteBackupService`, we can call this: `deleteAllBackupServices`

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/BackupConfigurationIT.kt
```diff
@@ -0,0 +1,110 @@
...
+                    val driver = browser.driver
+                    val webJira = WebJira(
+                        driver,
+                        jira.getUri(),
+                        user.password
+                    )
+                    goToServices(driver, jira)
+                    addBackupService(driver)
+
+                    TestingSetup(webJira, actionMeter).run()
```
dagguh: We can make it even simpler:
```suggestion
                    webJira.configureBackupPolicy().delete()
```
and nuke the `TestingSetup`

PR #27, opened 2019-08-12
Mark 3.8.0

dagguh:
https://packages.atlassian.com/maven-public/com/atlassian/performance/tools/jira-actions/3.8.0/


dagguh:
FYI @kema2000

PR #28, opened 2019-08-14
Mark 3.9.0 release

lipinskipawel:
https://packages.atlassian.com/maven-public/com/atlassian/performance/tools/jira-actions/3.9.0/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,10 @@ Adding a requirement of a major version of a dependency is breaking a contract.
 Dropping a requirement of a major version of a dependency is a new contract.
 
 ## [Unreleased]
-[Unreleased]: https://github.com/atlassian/jira-actions/compare/release-3.8.0...master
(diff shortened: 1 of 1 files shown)


lipinskipawel:
To resolve this problem I have seen 3 options available for us:
 - Increase the travis_wait in `.travis.yml` file
 - Delete the code which is testing BackupConfiguration class
 - Use the Travis Ci services like `docker` to provide chromdwriver and avoid OOM. 

First one is a workaround and maybe flaky in the future.
The second one is bad for many reasons so I don’t know if we should even consider this.
The third one, in my opinion, is the way to go. It’s the hardest but it will provide stability for not only this test but also for the future testing scenario.

Do we see any other solutions?
Which one solution do we want to choose?

dagguh:
So you reran it and it went green?

lipinskipawel:
Yes, but I believe that this problem will return.

PR #30, opened 2019-09-06
Add API for action metric parser format.

swidnow2:


src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/parser/format/MetricJsonFormat.kt
```diff
@@ -0,0 +1,10 @@
+package com.atlassian.performance.tools.jiraactions.api.parser.format
+
+import com.atlassian.performance.tools.jiraactions.api.ActionMetric
+import javax.json.JsonObject
+
+interface MetricJsonFormat {
+    fun serialize(actionMetric: ActionMetric): JsonObject
```
dagguh: This might look problematic, because in the future we might want o use more efficient future formats (like binary Protobuf), which are not JSON.
But we learned how to expand interfaces without breaking compatibility, so it's perfectly fine to leave this as is.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/parser/format/MetricCompactJsonFormat.kt
```diff
@@ -0,0 +1,43 @@
...
+import java.time.Duration
+import java.time.Instant
+import java.util.*
+import javax.json.Json
+import javax.json.JsonObject
+
+/**
+ * Format ignoring drilldown parameter
+ */
+class MetricCompactJsonFormat: MetricJsonFormat {
```
dagguh: Can we get add a test case to `ActionMetricsParserTest` just to see if this doesn't throw?

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/parser/ActionMetricsParserTest.kt
```diff
@@ -77,4 +79,51 @@ class ActionMetricsParserTest {
...
-}
\ No newline at end of file
+
+    @Test
+    fun shouldParseCompact() {
+        val metricsStream = javaClass.classLoader.getResourceAsStream("virtual-user-results/1/action-metrics.jpt")
+
+        val parser = ActionMetricsParser(MetricCompactJsonFormat())
+
+        @Suppress("Deprecation")
```
wyrzyk: Is not deprecated API available?

dagguh: Yeah, we can always `parser.stream(it).toList()`

dagguh:
There's a merge commit we gotta get rid of

dagguh:
Luckily, GitHub rebased it for us. No merge commit on master! 🎉

PR #33, opened 2019-12-20
avoid race condition in jquery initialisation

pbruski:


src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/DashboardPage.kt
```diff
@@ -46,6 +46,9 @@ class DashboardPage(
             //we currently support only single iframe on dashboard in this check
             return input.executeScript(
                 """
+                if (typeof $ === 'undefined') { //wait for jquery
+                    return false;
+                }
                 iframes = $('#dashboard').find('iframe');
```
pbruski: This causes about 5% failure rate of dashboard view in our tests

wyrzyk: Nice finding!

pbruski:
Rebased to remove merge commit.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/DashboardPage.kt
dagguh: Whoah!

PR #34, opened 2019-12-20
remove a-star query

pbruski:


src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/memories/adaptive/AdaptiveJqlMemory.kt
```diff
@@ -21,8 +21,7 @@ class AdaptiveJqlMemory(
     }
 
     private val jqls = mutableListOf(
-        TaggedBakedJql(BakedJql({ _ -> ""}, "resolved is not empty order by description"), BuiltInJQL.RESOLVED.name),
-        TaggedBakedJql(BakedJql({ _ -> ""}, "text ~ \"a*\" order by summary"), BuiltInJQL.GENERIC_WIDE.name)
```
pbruski: I don't think this query accurately reflect production use. Sure, it could happen now and then, but current set up seems to unfairly shift the performance bottleneck towards Lucene.
Ideally, we'd put realistic prod data here.

PR #35, opened 2019-12-20
replace n+1 remote calls with a single JS select

pbruski:


src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssueNavigatorPage.kt
```diff
@@ -36,11 +36,11 @@ class IssueNavigatorPage(
     }
 
     fun getIssueKeys(): Set<String> {
-        return driver
-            .findElements(By.className("issue-link-key"))
-            .map { it.text }
```
pbruski: each map invocation here means two remote calls

pbruski: this was found during VU profiling

src/main/kotlin/com/atlassian/performance/tools/jiraactions/webdriver/WebDriverUtils.kt
```diff
@@ -0,0 +1,48 @@
...
+import java.util.concurrent.atomic.AtomicInteger
+
+class WebDriverUtils {
+    companion object {
+        fun <T> executeScript(driver: WebDriver, js: String): T {
+            driver as JavascriptExecutor
+            return driver.executeScript(js) as T
+        }
+
+        fun sometimesApply(condition: ExpectedCondition<Boolean>): ExpectedCondition<Boolean> {
```
pbruski: I left the sometimes* methods here although they are technically used in a different PR. I struggle a bit with my workspace, there are too many interrelated changes there.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssueNavigatorPage.kt
dagguh: Did it help? What's the before and after?
I recall that the problem at hand is about VUs being the bottleneck after removing wildcard JQLs. Is that accurate?

src/main/kotlin/com/atlassian/performance/tools/jiraactions/webdriver/WebDriverUtils.kt
dagguh: Which PR is that?

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssueNavigatorPage.kt
pbruski: No, that's another problem. With this (and other related) changes, we are reaching 500 000 processed actions in the timespan of 20 minutes. The same tests without these changes were closer to 400 000 actions, all other things being the same.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/webdriver/WebDriverUtils.kt
pbruski: Not published yet, I want to get rid of all the other changes in my workspace first.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/webdriver/WebDriverUtils.kt
```diff
@@ -0,0 +1,48 @@
...
+            return object : ExpectedCondition<Boolean> {
+                val callCounter = AtomicInteger()
+
+                override fun apply(input: WebDriver?): Boolean {
+                    return callCounter.incrementAndGet() % 5 == 0 && condition.apply(input)!!
```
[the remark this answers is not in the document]
pbruski: No - this can be used in composite expressions.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/webdriver/WebDriverUtils.kt
```diff
@@ -0,0 +1,48 @@
...
+            }
+        }
+
+        fun sometimesLocate(condition: ExpectedCondition<WebElement>): ExpectedCondition<WebElement> {
+            return object : ExpectedCondition<WebElement> {
+                val callCounter = AtomicInteger()
+
+                override fun apply(input: WebDriver?) : WebElement {
+                    if (callCounter.incrementAndGet() % 5 != 0) {
+                        throw NoSuchElementException("We haven't actually looked for $condition")
```
dagguh:
```suggestion
                        return null
```

src/main/kotlin/com/atlassian/performance/tools/jiraactions/webdriver/WebDriverUtils.kt
dagguh: Note that this is not API and therefore won't be consumable outside of `jira-actions`. Unused internal code begs to be deleted. It's best to combine both new code and it's usages in the same commit. It's easier to understand in the future.

pbruski: Can I make it an API? I'd like to use it in software-actions too.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/webdriver/WebDriverUtils.kt
pbruski: This is modelled after presenceOfElementLocated, which throws or returns the element.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/webdriver/WebDriverUtils.kt
dagguh: Sure, gotta move it into an `api` subpackage

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/webdriver/WebDriverUtils.kt
```diff
@@ -0,0 +1,48 @@
+package com.atlassian.performance.tools.jiraactions.api.webdriver
+
+import org.openqa.selenium.JavascriptExecutor
+import org.openqa.selenium.NoSuchElementException
+import org.openqa.selenium.WebDriver
+import org.openqa.selenium.WebElement
+import org.openqa.selenium.support.ui.ExpectedCondition
+import java.util.concurrent.atomic.AtomicInteger
+
+class WebDriverUtils {
```
dagguh: Utils is a dangerous name to use, because it doesn't have a responsibility and attracts random code.
I usually warn about the future evolution of such classes, but we can already see it here:
-  a bit of syntax sugar around `WebDriver`
- `ExpectedCondition` wrappers, which don't interact with `WebDriver` instances at all

I'd move the syntax sugar to `WebDriverExtension` (a sin I committed a long time ago 😉).
I'd extract the "sometimes" to `ExtraExpectedConditions`/`LightweightConditions` or even promote each to a named implementation.

PS. I thought I posted this comment hours ago. Seems I didn't send it. Sorry for the feedback delay.

dagguh: Also, since it's API, I gotta ask about the `companion object`. Do you aim to provide Java `static`? In order to mimic `org.openqa.selenium.support.ui.ExpectedConditions` static factory methods? If so, we gotta use [`@JvmStatic`](https://kotlinlang.org/docs/reference/java-to-kotlin-interop.html#static-methods).
And if we don't care about the class name - we can use [package-level functions](https://kotlinlang.org/docs/reference/java-to-kotlin-interop.html#package-level-functions) without an encompassing class, trading `LightweightConditions.Companion.sometimesApply` for `LightweightConditionsKt.sometimesApply` in Java.

pbruski: Added

pbruski: > Utils is a dangerous name to use, because it doesn't have a responsibility and attracts random code.
> I usually warn about the future evolution of such classes, but we can already see it here:
> 
> * a bit of syntax sugar around `WebDriver`
> * `ExpectedCondition` wrappers, which don't interact with `WebDriver` instances at all
> 
Actually, all these interact with webdriver (which is the reason for the performance problem behind them).

> I'd move the syntax sugar to `WebDriverExtension` (a sin I committed a long time ago 😉).

Yeah, I saw that class. I don't really like that approach, I think it should only be applied to methods which very closely resemble the methods already available on that class. For example:
`
fun WebDriver.maybeFindElement(
    locator: By
): WebElement? {
    return this.findElements(locator).singleOrNull()
}
`
Would be OK, but the methods in this utils class wouldn't.

> I'd extract the "sometimes" to `ExtraExpectedConditions`/`LightweightConditions` or even promote each to a named implementation.

OK.

dagguh: > Actually, all these interact with webdriver (which is the reason for the performance problem behind them).

Oops, you're right, I missed this bit: `fun apply(input: WebDriver?)`.
OTOH it's still a good idea to find a more specific home and focus for the code. Otherwise most classes here (and in `virtual-users` and `jira-software-actions`) would be merged into a multi-KLOC `WebDriverUtils` god-object 😉

PR #37, opened 2020-01-04
a framework for expected conditions in pure javascript

pbruski:


pbruski:
Come to think of it, maybe putting it into separate library would be better.

dagguh:
> Come to think of it, maybe putting it into separate library would be better.

It's fine to incubate it here and when battle-tested extract outside.

pbruski:
Yeah, but I realised I want these in jira-software-actions too (for ViewBoardPage.waitForBoard and ViewBacklogPage.waitForBacklog). Also, this is something Confluence might be interested in too. 
I'll just make sure it's green with the jira-actions and extract it.

pbruski:
Closed in favour of https://github.com/atlassian/jira-actions/pull/38

PR #38, opened 2020-01-10
use native JavaScript conditions for performance-intensive checks

pbruski:


src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/WebDriverExtension.kt
```diff
@@ -25,6 +27,19 @@ fun <T> WebDriver.wait(
     ).until(condition)
 }
 
+@JvmOverloads
+fun WebDriver.wait(
+    timeout: Duration,
+    condition: NativeExpectedCondition,
```
dagguh: It appears on the API surface now, so it should be in Gradle `api` scope: https://ecosystem.atlassian.net/browse/JPERF-607

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/WebDriverExtension.kt
```diff
@@ -25,6 +27,19 @@ fun <T> WebDriver.wait(
...
+    precision: Duration = Duration.ofMillis(100)
```
dagguh: In the future, we should avoid Kotlin defaults in API.

PR #39, opened 2020-01-28
smarter action shuffler

pbruski:


pbruski:
Hold on. That should actually work per-node, right?

build.gradle.kts
```diff
@@ -43,6 +43,9 @@ dependencies {
     implementation("org.apache.commons:commons-math3:3.6.1")
     implementation("com.atlassian.performance.tools:concurrency:[1.0.0,2.0.0)")
     implementation("com.atlassian.performance:selenium-js:[1.0.0,2.0.0)")
+
+    runtimeOnly("org.jetbrains.kotlin:kotlin-reflect:${kotlinVersion}")
```
dagguh: Uh oh, reflection = danger zone. We're bypassing a lot of compilation-time guarantees.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/ActionShuffler.kt
```diff
@@ -0,0 +1,81 @@
...
+            return jql
+        }
+
+        override fun remember(memories: Collection<String>) {
+            throw UnsupportedOperationException()
+        }
+    }
+
+    companion object {
+        fun createRandomisedScenario(seededRandom: SeededRandom, actionProportions: Map<Action, Int>,
```
dagguh: What's the problem we're trying to solve here?

Do skipped actions malform the `ActionMetric.label` proportions too much? Our initial workaround was to run the test for a longer time so they get averaged out.
If that is not effective enough, then we can avoid action skips by using a self-loading `Memory`, e.g. [pass `this` to an action during `Memory.recall`](https://bitbucket.org/atlassian/atlassian-audit/pull-requests/218#Latlassian-audit-performance-tests/src/main/java/com/atlassian/audit/performance/jira/vu/memory/DynamicIssueKeyMemory.javaT42):
```kotlin
fun recall(): String {
    if (issueKeys.isEmpty()) {
        SearchJqlAction(jira, throwawayActionMeter, IssueJqlMemory(), this).run()
    }
}
```
This way you can avoid skips with local logic rather than a global order-controlling scheme.

pbruski: Yes. I want to preserve the requested proportions. During our runs, ~1500 view issue calls are skipped. This looks like a minimally invasive way of achieving this, without altering the way scenarios are created.
The problem with hiding the query inside a different action is that you lose control over proportions - if there are permission issues, the JQL action can be executed as often as view issue, which we don't want. If we stick to sorting, everything is transparent and there are no surprises hiding in the implicit initialisation.

pbruski: The same problem affects ViewBoard, btw.

pbruski: Let's try a different question: is there any benefit to keeping a fully-randomised shuffle in the scenario?

dagguh: 🤔 
I'll come back to this next week

PR #41, opened 2020-01-29
upgrade docker-infrastructure

pbruski:


src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/RichTextEditorIT.kt
```diff
@@ -60,7 +61,8 @@ class RichTextEditorIT {
             .build()
             .provision()
             .use { jira ->
-                DockerisedChrome().start().use { browser ->
+                val recordings = Paths.get("build/diagnoses/recordings/" + this::class.java.simpleName)
+                DockerisedChrome(recordings).start().use { browser ->
```
dagguh: Woop woop! Sounds like a great way to improve the local dev loop 👍

PR #42, opened 2020-01-29
move popup dismissal to login, dismiss more popups

pbruski:


src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/action/SetUpAction.kt
```diff
@@ -3,13 +3,27 @@ package com.atlassian.performance.tools.jiraactions.api.action
...
+import org.openqa.selenium.By
 
 class SetUpAction(
     private val jira: WebJira,
     private val meter: ActionMeter
 ) : Action {
     override fun run() {
         meter.measure(SET_UP) {
+
+            val popupDismissalLocators = listOf(
```
pbruski: or should that be in WebJira?

pbruski: or maybe there should be a LoginWithPopupCleanup action?

dagguh: The popup interactions themselves should be extracted to a [page object](https://github.com/SeleniumHQ/selenium/wiki/PageObjects), existing or new.
The call to `dismissPopups` can be included in the default login action. `LogInWithoutPopups` also sounds good.

`SetUpAction` is used once per test by an "admin user", which can be different from the individual VU users.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/DashboardPage.kt
```diff
@@ -14,10 +14,17 @@ class DashboardPage(
...
+            By.id("nps-acknowledgement-accept-button"),
+            By.cssSelector(".jira-help-tip .cancel"),
+            By.className("postsetup-close-link")
+        )
+        
+        for (locator in popupDismissalLocators) {
+            driver.findElements(locator)
+                .filter { it.isEnabled && it.isDisplayed }
+                .forEach { it.click() }
+        }
```
dagguh: If you want to keep chaining:
```kotlin
listOf(
    By.id("nps-acknowledgement-accept-button"),
    By.cssSelector(".jira-help-tip .cancel"),
    By.className("postsetup-close-link")
)
    .flatMap { driver.findElements(it) }
    .filter { it.isEnabled && it.isDisplayed }
    .forEach { it.click() }
```

pbruski: I don't, really - there may be an overlap between these selectors, for example AUI flags can be handled by the "close" button but also "Don't remind me again". If we run the second first, the first one won't be interactable anymore and will result in an exception.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/action/SetUpAction.kt
pbruski: It's not really a page object, because it's not a page... I think I'll add it to the login action, but run it only once per user.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/DashboardPage.kt
dagguh: `listOf().asSequence()` should do the trick

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/action/LogInAction.kt
```diff
@@ -4,17 +4,38 @@ import com.atlassian.performance.tools.jiraactions.api.LOG_IN
...
 import com.atlassian.performance.tools.jiraactions.api.memories.UserMemory
+import org.openqa.selenium.By
+import java.util.concurrent.ConcurrentHashMap
 
 class LogInAction(
     private val jira: WebJira,
     private val meter: ActionMeter,
     private val userMemory: UserMemory
 ) : Action {
+    private val loggedInUsers = ConcurrentHashMap.newKeySet<String>()
```
dagguh: Each VU has it's own copy of `LogInAction`, or more specifically, each calls `Scenario.getLogInAction` - they don't talk to each other by default. The `Scenario` impl can choose to return new copy each time or a shared one. By default `Scenario` creates new copies.
So right now this state is not useful, it will contain zero or one elements.

Also, instead of making VUs coordinate with one another, maybe we could run `dismissPopups` regardless? It's idempotent, right?

pbruski: So maybe I should make it static? It is idempotent, and I wouldn't have an issue if this was a functional test. But for a performance test I would prefer if we didn't do (too many) superfluous actions.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/action/SetUpAction.kt
dagguh: `SingleSelect` also isn't a page, but it represents a part of a page.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/action/LogInAction.kt
dagguh: Sounds like premature optimization. How often do we log in? How expensive is the dismiss code?

pbruski: Yeah. Maybe. I think Selenium usage requires a bit of a mindset change. Premature optimization avoidance is good for places where we have perfomance tests. Nobody tests the performance of the performance tool :-)

pbruski: I mean, I can remove it if you want. It's just that I think we should really be very careful whenever we interact with RemoteWebElements in JPT code as a rule.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/action/SetUpAction.kt
pbruski: Fair point. I can extract it. Should it be AnyPageWithPopups ? :-)

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/action/LogInAction.kt
dagguh: OTOH we should also avoid unnecessary locks across threads, which were generally thread-confined so far. And each piece of additional complexity risks bugs, left to be discovered in a few months.

> Nobody tests the performance of the performance tool :-)

Well we could 😛 At least from time to time.
Perhaps we could reuse `com.atlassian.performance.tools.infrastructure.api.profiler.AsyncProfiler` around `SshVirtualUsers`. It'd require `jira-actions` tests to switch from `docker-infrastructure` `infrastructure` + `ssh-ubuntu`.

From a pragmatic point of view:
* this optimization works only when VUs share logins (e.g. not for `RestUserGenerator`)
* it reduces `dismissPopups` calls by VU count minus VU node count
* the biggest tests would save ~1000 such calls, but they do use `RestUserGenerator` or similar
* in medium/small tests it would save ~20 such calls

IMHO the savings are not worth the complexity

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/measure/ActionMeterTest.kt
```diff
@@ -83,7 +83,7 @@ class ActionMeterTest {
         )
 
         try {
-            actionMeter.measure(CREATE_ISSUE) { throw Exception("oops") }
+            actionMeter.measure(CREATE_ISSUE) { throw Exception("Ignore this exception, it's test-only") }
```
dagguh: Sorry for the WTF factor you must have went through here

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/action/LogInAction.kt
pbruski: removed

src/main/kotlin/com/atlassian/performance/tools/jiraactions/page/NotificationPopUps.kt
```diff
@@ -0,0 +1,40 @@
+package com.atlassian.performance.tools.jiraactions.page
+
+import org.openqa.selenium.By
+import org.openqa.selenium.WebDriver
+
+internal class NotificationPopUps(private val driver: WebDriver) {
```
dagguh: Nice one!

build.gradle.kts
```diff
@@ -84,6 +84,11 @@ tasks.getByName("test", Test::class).apply {
 }
 
 val testIntegration = task<Test>("testIntegration") {
+    testLogging.exceptionFormat =  org.gradle.api.tasks.testing.logging.TestExceptionFormat.FULL
+    testLogging.showStackTraces = true
+    testLogging.showExceptions = true
+    testLogging.showCauses = true
+    testLogging.showStandardStreams = true
```
dagguh: Hacks for Travis :)

src/main/kotlin/com/atlassian/performance/tools/jiraactions/page/NotificationPopUps.kt
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    private fun clickAll(locator: By): NotificationPopUps {
+        driver.findElements(locator)
+            .filter { it.isEnabled && it.isDisplayed }
+            .forEach { 
+                try {
+                    it.click()
+                } catch (e: Exception) {
+                    println((driver as RemoteWebDriver).getScreenshotAs(OutputType.BASE64))
```
dagguh: Travis really hurts 😬
Could you mark these "artifacts via stdout" workarounds in comments or as a separate commit?

src/main/kotlin/com/atlassian/performance/tools/jiraactions/page/IssueCreateDialog.kt
```diff
@@ -6,7 +6,9 @@ import com.atlassian.performance.tools.jiraactions.api.page.isElementPresent
 import com.atlassian.performance.tools.jiraactions.api.page.tolerateDirtyFormsOnCurrentPage
 import com.atlassian.performance.tools.jiraactions.api.page.wait
 import org.openqa.selenium.By
+import org.openqa.selenium.OutputType
 import org.openqa.selenium.WebDriver
+import org.openqa.selenium.remote.RemoteWebDriver
```
dagguh: New imports without usages?

build.gradle.kts
pbruski: Not really, I think these are useful on their own.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/page/IssueCreateDialog.kt
pbruski: removing

src/main/kotlin/com/atlassian/performance/tools/jiraactions/page/NotificationPopUps.kt
pbruski: I'll just remove it, let's hope they won't be needed 'anymore'.

build.gradle.kts
dagguh: Gradle output should be nearly empty. Only actionables should pop up, e.g. warnings, errors.

See [this green build](https://travis-ci.com/atlassian/jira-actions/jobs/281831631). The stdout is full of noise, because some tests cover the unhappy paths. If someone cares about outputs of a test, they can take a look at the test report. The entire stdout, stderr and stacktrace are not necessary, because the test report has it all. And test reports are much better structured than a text stream on stdout.

PR #44, opened 2020-02-03
update-to-the-latest-Selenium-3 v2

pbruski:


src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/BackupConfiguration.kt
```diff
@@ -33,8 +35,13 @@ class BackupConfiguration(
...
 
     private fun deleteBackupService() {
-        driver.findElement(deleteBackupLocator).click()
-        driver.switchTo().alert().accept()
+        driver.wait(ExpectedConditions.elementToBeClickable(deleteBackupLocator)).click()
+        try {
+            driver.wait(ExpectedConditions.alertIsPresent())
+            driver.switchTo().alert().accept()
+        } catch (e: TimeoutException) {
+            //this alert sometimes doesn't show...
```
dagguh: Oh JIra... 😉

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/webdriver/JavaScriptUtils.kt
```diff
@@ -5,11 +5,10 @@ import org.openqa.selenium.WebDriver
 
 class JavaScriptUtils {
     companion object {
-
         @JvmStatic
-        fun <T> executeScript(driver: WebDriver, js: String): T {
+        fun <T> executeScript(driver: WebDriver, js: String, vararg args: Any): T {
```
dagguh: It's ok to change this API, because it was not released yet:
`git tag --contains ba5c8224691e748c6cf74a697a651c4002724fad`

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/webdriver/JavaScriptUtils.kt
```diff
@@ -5,11 +5,10 @@ import org.openqa.selenium.WebDriver
...
             driver as JavascriptExecutor
-            return driver.executeScript(js) as T
+            return driver.executeScript(js, *args) as T
```
dagguh: Unchecked cast warning

src/main/kotlin/com/atlassian/performance/tools/jiraactions/page/NotificationPopUps.kt
```diff
@@ -33,8 +66,11 @@ internal class NotificationPopUps(private val driver: WebDriver) {
 
     private fun clickAll(locator: By): NotificationPopUps {
         driver.findElements(locator)
-            .filter { it.isEnabled && it.isDisplayed }
-            .forEach { it.click() }
+            .forEach {
+                // this lets us click flags that are hidden behind other flags
+                // otherwise we would have to wait for each flag to slide away
```
dagguh: I'll let it slide 😉

src/main/kotlin/com/atlassian/performance/tools/jiraactions/page/SingleSelect.kt
```diff
@@ -22,28 +23,31 @@ internal class SingleSelect(
...
+        val suggestions = driver
             .wait(
                 presenceOfElementLocated(By.cssSelector(".ajs-layer.active"))
             )
             .findElements(By.className("aui-list-item-link"))
             .map { it.text }
+        
+        // we now have to restore the select to it's usual state,
+        // otherwise we will mess up 'select-all-on-click' behaviour
+        // We click to hide the menu and tab to get out of the input element
```
dagguh: Interesting, good find

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/JiraCoreScenarioIT.kt
```diff
@@ -127,9 +130,31 @@ class JiraCoreScenarioIT {
...
+        sendKeys(driver, By.id("serviceName"), "another backup")
+        sendKeys(driver, By.id("serviceClass"), "com.atlassian.jira.service.services.export.ExportService")
+        
         driver.findElementById("addservice_submit").click()
-        driver.findElementById("update_submit").click()
+
+        try {
+            driver.findElementById("update_submit").click()
+        } catch (e: Exception) {
+            println((driver as RemoteWebDriver).getScreenshotAs(OutputType.BASE64))
```
dagguh: Let's:
- [ ] explain this is only for TravisCI

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/JiraCoreScenarioIT.kt
```diff
@@ -127,9 +130,31 @@ class JiraCoreScenarioIT {
...
+            throw e
+        }
+    }
+
+    private fun sendKeys(driver: RemoteWebDriver, locator: By, text: String) {
+        val element = driver.wait(ExpectedConditions.elementToBeClickable(locator))
+        element.click()
+
+        // It's hard to say when the keys can be sent. They seem to randomly get lost.
+        for (i in 0..10) {
```
dagguh:
```suggestion
        repeat(10) {
```

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/webdriver/JavaScriptUtils.kt
pbruski: Hmmm... I don't think it would be possible in Java, is it possible in Kotlin?

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/JiraCoreScenarioIT.kt
pbruski: nice!

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/webdriver/JavaScriptUtils.kt
dagguh: https://kotlinlang.org/docs/reference/typecasts.html#safe-nullable-cast-operator

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/JiraCoreScenarioIT.kt
pbruski: Hmm, then we won't be able to break...

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/webdriver/JavaScriptUtils.kt
dagguh: Oops, the safe cast does not work on generics due to type erasure. So we can remove generics by creating a method for each of:
`@return One of Boolean, Long, Double, String, List, Map or WebElement.`

If that's too much, then at least let's suppress it.

pbruski: The other thing is that I think we WANT to have an exception thrown if the JS returned a value of an unexpected type.

pbruski: suppressed.

dagguh: Yeah, that would be the part of transforming the `T?` local value into the `T` return value, ie.
```kotlin
val result = driver.executeScript(js) as? T
return result ?: throw Exception("Unexpected return type: ${result.javaClass}")
```
Which would basically reimplement the `ClassCastException`, so you're right, no point in that.
Let's just suppress it.

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/JiraCoreScenarioIT.kt
dagguh: I missed the `break`, nevermind me 🙂

PR #46, opened 2020-02-06
a more reliable value setter for single select

pbruski:


src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/action/CreateIssueAction.kt
```diff
@@ -31,11 +31,11 @@ class CreateIssueAction(
                 .waitForDialog()
                 .showAllFields()
                 .selectProject(project.name)
-                .selectIssueType(
+                .selectIssueType { issueTypes ->
                     seededRandom.pick(
-                        issueCreateDialog.getIssueTypes().filter { it != "Epic" }
+                        issueTypes.filter { it != "Epic" }
```
dagguh: FYI the filtering is no longer necessary. It was done to avoid filling in a mandatory `Epic Link` field a loooong time ago.
Now we fill in all required fields automatically, a few lines below: `issueCreateDialog.fillRequiredFields()`.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/page/SingleSelect.kt
```diff
@@ -15,16 +16,24 @@ internal class SingleSelect(
...
         val inputElement = driver.wait(
             timeout = Duration.ofSeconds(8),
-            condition = elementToBeClickable(input)
+            condition = elementToBeClickable(inputLocator)
         )
         inputElement.click()
         inputElement.sendKeys(Keys.BACK_SPACE, value, Keys.TAB)
     }
+
+    fun select(picker: (Collection<String>) -> String) {
```
dagguh: Thanks for the hardening. Could you add it to the changelog?

- [ ] mention hardening in the changelog

Also, it'd be nice to have a JPERF ticket, where we could describe the symptom. This way when someone else inevitably encounters the same problem, they'll know it's fixed and where to bump to.
It's enough for the bug description to contain the stacktrace.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/action/CreateIssueAction.kt
pbruski: OK, removing then.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/page/SingleSelect.kt
pbruski: It's actually a better fix for https://ecosystem.atlassian.net/browse/JPERF-298

dagguh: Noice, let's mention it's a fix for this ticket. The changelog diff is still empty.

pbruski: It's already in the changelog :-)

dagguh: Ooooh, sorry. I'm not used to [a long-running "Unreleased" section](https://github.com/atlassian/jira-actions/compare/release-3.10.0...master).

PR #51, opened 2020-02-10
Mark 3.10.2

pbruski:


CHANGELOG.md
```diff
@@ -24,9 +24,12 @@ Adding a requirement of a major version of a dependency is breaking a contract.
 Dropping a requirement of a major version of a dependency is a new contract.
 
 ## [Unreleased]
-[Unreleased]: https://github.com/atlassian/jira-actions/compare/release-3.10.1...master
+[Unreleased]: https://github.com/atlassian/jira-actions/compare/release-3.10.2...master
+
+## [3.10.2] - 2020-02-11
```
dagguh: It was not released yet: https://packages.atlassian.com/maven-public/com/atlassian/performance/tools/jira-actions/
The GitHub URLs are not testable.

We learned the hard way to mark releases **after** they happen. In other words, write only true things, not things we wish were true.

Sometimes the release doesn't work and gets postponed, sometimes for days, sometimes for weeks. Sometimes we don't have control. Sometimes it's not 3.10.2, but turns out it's 3.11.0.
Then we have to edit changelog again.

- [ ] recheck and merge after the release

dagguh: Confirmed: https://packages.atlassian.com/maven-public/com/atlassian/performance/tools/jira-actions/3.10.2/ 👍

PR #59, opened 2020-03-17
ActionMetric copy builder, ActionMeter builder, ActionMeter drrilldown condition

Mac-Ste:
- Added copy builder constructor to ActionMetric.Builder
- Added ActionMeter builder and deprecated `withX` sytnax
- Added drilldown custom condition based on ActionMetric to ActionMeter

### DIFF
--- a/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/ActionMetric.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/ActionMetric.kt
@@ -77,8 +77,8 @@ data class ActionMetric @Deprecated("Use ActionMetric.Builder instead.") constru
 
     class Builder(
         private val label: String,
-        private val result: ActionResult,
-        private val duration: Duration,
+        private var result: ActionResult,
+        private var duration: Duration,
         private val start: Instant
     ) {
         private var observation: JsonObject? = null
@@ -86,20 +86,20 @@ data class ActionMetric @Deprecated("Use ActionMetric.Builder instead.") constru
         private var virtualUser: UUID = UUID.randomUUID()
 
 
-        constructor(actionMetric: ActionMetric) : this(
-            actionMetric.label,
-            actionMetric.result,
-            actionMetric.duration,
-            actionMetric.start
+        constructor(label: String, virtualUser: UUID, start: Instant) : this(
+            label = label,
+            result = ActionResult.OK,
+            duration = Duration.ofMillis(0),
+            start = start
         ){
-            observation = actionMetric.observation
-            drilldown = actionMetric.drilldown
-            virtualUser = actionMetric.virtualUser
+            this.virtualUser = virtualUser
         }
 
         fun observation(observation: JsonObject?) = apply { this.observation = observation }
         fun drilldown(drilldown: RecordedPerformanceEntries?) = apply { this.drilldown = drilldown }
         fun virtualUser(virtualUser: UUID) = apply { this.virtualUser = virtualUser }
+        fun result(result: ActionResult) = apply { this.result = result }
+        fun duration(duration: Duration) = apply { this.duration = duration }
 
         @Suppress("DEPRECATION")
         fun build() = ActionMetric(
--- a/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/ActionMetric.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/ActionMetric.kt
@@ -85,6 +85,18 @@ data class ActionMetric @Deprecated("Use ActionMetric.Builder instead.") constru
         private var drilldown: RecordedPerformanceEntries? = null
         private var virtualUser: UUID = UUID.randomUUID()
 
+
+        constructor(actionMetric: ActionMetric) : this(
+            actionMetric.label,
+            actionMetric.result,
+            actionMetric.duration,
+            actionMetric.start
+        ){
+            observation = actionMetric.observation
+            drilldown = actionMetric.drilldown
+            virtualUser = actionMetric.virtualUser
+        }
+
         fun observation(observation: JsonObject?) = apply { this.observation = observation }
         fun drilldown(drilldown: RecordedPerformanceEntries?) = apply { this.drilldown = drilldown }
         fun virtualUser(virtualUser: UUID) = apply { this.virtualUser = virtualUser }
--- a/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/measure/ActionMeter.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/measure/ActionMeter.kt
@@ -30,18 +30,24 @@ class ActionMeter private constructor(
     private val clock: Clock,
     private val w3cPerformanceTimeline: W3cPerformanceTimeline,
     private val drilldownCondition: Predicate<ActionMetric>
-)
-{
+) {
 
     @Deprecated(
-        message = "Use ActionMeter.Builder instead"
+        message = "Use ActionMeter.Builder instead",
+        replaceWith = ReplaceWith("ActionMeter.Builder(" +
+            "\nvirtualUser = virtualUser," +
+            "\noutput = output" +
+            "\n)" +
+            "\n.clock(clock)" +
+            "\n.performanceTimeline(w3cPerformanceTimeline)" +
+            "\n.build()")
     )
     constructor(
         virtualUser: UUID,
         output: ActionMetricOutput,
         clock: Clock,
         w3cPerformanceTimeline: W3cPerformanceTimeline
-    ) : this (
+    ) : this(
         virtualUser = virtualUser,
         output = output,
         clock = clock,
@@ -50,7 +56,13 @@ class ActionMeter private constructor(
     )
 
     @Deprecated(
-        message = "Use ActionMeter.builder instead"
+        message = "Use ActionMeter.Builder instead",
+        replaceWith = ReplaceWith("ActionMeter.Builder(" +
+            "\nvirtualUser = virtualUser," +
+            "\noutput = output" +
+            "\n)" +
+            "\n.clock(clock)" +
+            "\n.build()")
     )
     constructor(
         virtualUser: UUID,
@@ -61,7 +73,7 @@ class ActionMeter private constructor(
         output = output,
         clock = clock,
         w3cPerformanceTimeline = DisabledW3cPerformanceTimeline(),
-        drilldownCondition = Predicate { false }
+        drilldownCondition = Predicate { true }
     )
 
     /**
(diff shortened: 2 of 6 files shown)


src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/measure/ActionMeter.kt
```diff
@@ -76,50 +97,80 @@ class ActionMeter(
         observation: (T) -> JsonObject?
     ): T {
         val start = clock.instant()
+        var actionResult: ActionResult = ActionResult.OK
```
dagguh: We avoid `var`s whenever possible. Tracking mutable state is error-prone. Builders already encapsulate mutability.
So I'd start with an `ActionMetric.Builder`, filling in the common fields, like `label`, `start` and `virtualUser` and let the rest of the control flow contribute to the state.

- [ ] avoid non-builder `var`s

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/measure/ActionMeter.kt
```diff
@@ -76,50 +97,80 @@ class ActionMeter(
...
                     .drilldown(w3cPerformanceTimeline.record())
                     .build()
-            )
-            throw Exception("Action '${key.label}' $result", e)
+            }
+            output.write(metric)
         }
     }
 
+    @Deprecated( message = "Use builder instead")
```
dagguh: Pro-tip, you can use [`ReplaceWith`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-replace-with/) to provide a top-notch migration experience.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/measure/ActionMeter.kt
```diff
@@ -76,50 +97,80 @@ class ActionMeter(
...
         clock = clock,
-        w3cPerformanceTimeline = w3cPerformanceTimeline
+        w3cPerformanceTimeline = w3cPerformanceTimeline,
+        drilldownCondition = Predicate { true }
     )
+
+    class Builder(
+        private val virtualUser: UUID,
+        private val output: ActionMetricOutput,
+        private val clock: Clock
```
dagguh: Builders are most convenient if they don't require too many params. So we could provide a default clock: `Clock.systemUTC()`

- [ ] don't require a clock

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/measure/ActionMeter.kt
```diff
@@ -76,50 +97,80 @@ class ActionMeter(
...
+    ){
+        private var w3cPerformanceTimeline: W3cPerformanceTimeline = DisabledW3cPerformanceTimeline()
+        private var drilldownCondition: Predicate<ActionMetric> = Predicate { true }
+
+        constructor(meter: ActionMeter): this(
```
dagguh: With the `var` avoidance, we might not even need a copying builder here.

Mac-Ste: not sure, if I follow. You mean instead of:
`ActionMeter.Builder(meter)`
invoking it as:
`ActionMeter.Builder(meter.virtualUser, meter.output)` ?
The one before seems more clear, or you mean reusing old builder to build new copy?

dagguh: Oops, you're right, this one is ok. I meant the `ActionMetric` copying builder. Imma comment there

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/ActionMetric.kt
```diff
@@ -85,6 +85,18 @@ data class ActionMetric @Deprecated("Use ActionMetric.Builder instead.") constru
         private var drilldown: RecordedPerformanceEntries? = null
         private var virtualUser: UUID = UUID.randomUUID()
 
+
+        constructor(actionMetric: ActionMetric) : this(
```
dagguh: If we [avoid `var`s](https://github.com/atlassian/jira-actions/pull/59#discussion_r393700108), we won't need to copy `ActionMetric`s.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/measure/ActionMeter.kt
```diff
@@ -21,16 +22,35 @@ import javax.json.JsonObject
...
     private val output: ActionMetricOutput,
     private val clock: Clock,
-    private val w3cPerformanceTimeline: W3cPerformanceTimeline
-) {
+    private val w3cPerformanceTimeline: W3cPerformanceTimeline,
+    private val drilldownCondition: Predicate<ActionMetric>
+)
+{
+
+    @Deprecated(
```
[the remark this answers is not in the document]
Mac-Ste: Changed deprecated code in tests to use builder

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/measure/ActionMeter.kt
```diff
@@ -50,7 +56,13 @@ class ActionMeter private constructor(
     )
 
     @Deprecated(
-        message = "Use ActionMeter.builder instead"
+        message = "Use ActionMeter.Builder instead",
+        replaceWith = ReplaceWith("ActionMeter.Builder(" +
```
Mac-Ste: Not sure if I should include formatting here, but if I put it into one line IntelliJ was doing some weird formatting. (kotlin multiline string did not work here)

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/measure/ActionMeter.kt
```diff
@@ -97,42 +109,43 @@ class ActionMeter private constructor(
...
-                result = actionResult,
-                start = start,
-                duration = Duration.between(start, clock.instant())
+            actionMetricBuilder.duration(Duration.between(start, clock.instant()))
+
+            val metricWithoutDrilldown = actionMetricBuilder.build()
+
+            output.write(
+                if (drilldownCondition.test(metricWithoutDrilldown)) {
+                    actionMetricBuilder
```
Mac-Ste: No sure if it is ok to reuse builder tbh, but couldn't think of better solution there WDYT?

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/measure/ActionMeter.kt
dagguh: Nice, you went the extra mile. I'd be fine with a single long line. This is about the content of the replacement. We can leave the callers code style to the callers IDE.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/measure/ActionMeter.kt
dagguh: Reusing the builder is the right pattern. Builders let you "build objects up" in portions by design.
You can avoid `val metricWithoutDrilldown = actionMetricBuilder.build()` entirely. We don't have to rely on the return value from `drilldown`. Builders return `this` purely for syntactic sugar.

- [ ] avoid `metricWithoutDrilldown`

Mac-Ste: Well we rely on metric build by the builder to know if drilldown should be applied or not. The only way I could think of to get rid of `metricWithoutDrilldown` is by using `let` on metric. Pushed the changes right now

dagguh: Aaaah, right, it's the argument for the condition. Makes sense 👍

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/ActionMetric.kt
```diff
@@ -77,17 +77,29 @@ data class ActionMetric @Deprecated("Use ActionMetric.Builder instead.") constru
...
+        private var result: ActionResult,
+        private var duration: Duration,
         private val start: Instant
     ) {
         private var observation: JsonObject? = null
         private var drilldown: RecordedPerformanceEntries? = null
         private var virtualUser: UUID = UUID.randomUUID()
 
+
+        constructor(label: String, virtualUser: UUID, start: Instant) : this(
```
dagguh: Ad hoc constructors in Builders are suspicious. Builders hold defaults of all optional fields and enforce passing the required fields. Every other permutation can be reduced to calling the required constructor and mutators.
Builders only need the required-params (no good defaults) constructor and a copying constructor.

- [ ] avoid ad hoc builder constructors

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/measure/ActionMeter.kt
```diff
@@ -76,50 +109,82 @@ class ActionMeter(
...
+                    } else {
+                        it
+                    }
+                }
             )
-            throw Exception("Action '${key.label}' $result", e)
         }
     }
 
+    @Deprecated(message = "Use ActionMeter.Builder instead")
```
dagguh: `ReplaceWith` should be easy enough here. I don't require it though.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/measure/ActionMeter.kt
```diff
@@ -76,50 +109,82 @@ class ActionMeter(
...
         virtualUser = virtualUser,
         output = output,
         clock = clock,
-        w3cPerformanceTimeline = w3cPerformanceTimeline
+        w3cPerformanceTimeline = w3cPerformanceTimeline,
+        drilldownCondition = Predicate { true }
     )
+
+    class Builder(
+        private val virtualUser: UUID,
```
dagguh: There's a good default: `UUID.randomUUID`

- [ ] default for `virtualUser: UUID`

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/measure/ActionMeter.kt
```diff
@@ -76,50 +109,82 @@ class ActionMeter(
...
+        private val output: ActionMetricOutput
```
dagguh: A default of `ThrowawayActionMetricOutput` could work. Such a throwaway meter is useful sometimes, but I guess it could be a big WTF factor if by default the meter would drop measurements. So I agree, the choice of `output` should be forced on the caller 👍

dagguh:
Continued in https://github.com/atlassian/jira-actions/pull/60

PR #60, opened 2020-04-10
Added ActionMetric copy builder, ActionMeter builder, ActionMeter dri…

wyrzyk:
I based it on Maciej's PR (https://github.com/atlassian/jira-actions/pull/59). Fixed some issues + added a test to cover the new functionality.


src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/measure/ActionMeter.kt
```diff
@@ -76,50 +109,83 @@ class ActionMeter(
         observation: (T) -> JsonObject?
     ): T {
         val start = clock.instant()
+        val actionMetricBuilder = ActionMetric.Builder(
```
dagguh, line 112: This uses dangerous defaults like `result: OK` and `duration: ZERO`. Perhaps we could avoid fake values only to hope they'll be mutated later?
The hardship comes from the fact that we're building an `ActionMetric` for both the `try` block and `catch` block at the same time. It might be worth sacrificing this forced reuse to gain safety.

So we gotta do one of the following:
* avoid dangerous defaults in the builder, stop reusing the builder for both happy and unhappy paths
* accept dangerous defaults and make them actual builder defaults instead of ad hoc secondary builder constructor

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/ActionMetric.kt
```diff
@@ -77,17 +77,29 @@ data class ActionMetric @Deprecated("Use ActionMetric.Builder instead.") constru
...
+        private var result: ActionResult,
+        private var duration: Duration,
         private val start: Instant
     ) {
         private var observation: JsonObject? = null
         private var drilldown: RecordedPerformanceEntries? = null
         private var virtualUser: UUID = UUID.randomUUID()
 
+
+        constructor(label: String, virtualUser: UUID, start: Instant) : this(
```
dagguh, line 89: This is the "dangerous defaults ad hoc constructor" mentioned later

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/measure/ActionMeterTest.kt
```diff
@@ -98,6 +104,34 @@ class ActionMeterTest {
             .isNotNull
     }
 
+    @Test
+    fun shouldDrilldownConditionWork() {
```
dagguh, line 108: Grammar nit:
> ActionMeter~Test~ should drilldown condition work 🥴 

```suggestion
    fun shouldRespectDrilldownCondition() {
```

> ActionMeter~Test~ should respect drilldown condition 🤓

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/measure/ActionMeterTest.kt
```diff
@@ -98,6 +104,34 @@ class ActionMeterTest {
...
+        val tick = ofSeconds(1)
+        val clock = TickingClock(start, tick)
+        val output = CollectionActionMetricOutput(mutableListOf())
+        val w3cPerformanceTimelineMock = HardcodedTimeline(RecordedPerformanceEntries(emptyList(), emptyList()))
+        val actionMeter = ActionMeter.Builder(
+            output = output
+        )
+            .drilldownCondition(Predicate { actionMetric -> CREATE_ISSUE.label == actionMetric.label })
```
dagguh, line 116: Nice test, clean and simple.
Can we make it a red test first? We can add the new API, but the SUT can ignore it (not call `predicate.test`) - that should make it red

- [ ] red-green pls

wyrzyk: I don't have plans to make it red-green. I'd just wanna finish the PR to unlock the work I wanted to work on.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/ActionMetric.kt
wyrzyk: I overlooked that. Thanks.

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/measure/ActionMeterTest.kt
dagguh: Everyone is always in a hurry. And then we complain about the quality, when we're in a hurry again in the future.
I'll do it

dagguh: You can `git reset --hard` onto https://github.com/dagguh/jira-actions/tree/issue/JCES-1334-drilldown-condition

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/measure/ActionMeter.kt
```diff
@@ -76,50 +109,83 @@ class ActionMeter(
...
-            val result = if (e.representsInterrupt()) {
+            val actionResult = if (e.representsInterrupt()) {
                 ActionResult.INTERRUPTED
             } else {
                 ActionResult.ERROR
             }
+            actionMetricBuilder.result(actionResult)
+            throw Exception("Action '${key.label}' $actionResult", e)
+        } finally {
+            actionMetricBuilder.duration(Duration.between(start, clock.instant()))
```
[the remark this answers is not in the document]
wyrzyk: Done

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/measure/ActionMeterTest.kt
wyrzyk: done

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/measure/ActionMeterTest.kt
wyrzyk: So the first commit adds the functionality with a bug, and another fixes it. I would not say it increases the quality.

wyrzyk: BTW working on a PR in parallel doesn't bring much value. The time needed to resolve conflicts probably would be greater than the help with the change.

dagguh: Ok, so adding a new feature and test for it in red-green fashion apparently "feels weird". Maybe so.
I won't push it further. I'll try to do the TDD red-first even for new features myself. Maybe it'll be more convincing.

dagguh:
Nit: the commit message is out of date, we no longer add "ActionMetric copy builder". Also it was too long.
I'd suggest:
> Add `ActionMeter` drilldown condition and its builder

- [ ] update commit message

Adding it as a task to consider, but I leave the call to you

PR #65, opened 2020-04-16
Refactor tests to create a fast local dev loop

pbruski:


src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/JiraCoreScenarioDevloop.kt
```diff
@@ -0,0 +1,54 @@
...
+import org.openqa.selenium.chrome.ChromeOptions
+import org.openqa.selenium.logging.LogType
+import org.openqa.selenium.logging.LoggingPreferences
+import org.openqa.selenium.remote.CapabilityType
+import org.openqa.selenium.remote.RemoteWebDriver
+import java.net.URI
+import java.util.logging.Level
+
+/**
+ * This class is meant to be used in a local devloop, where JIra is already running.and Chrome is locally available.
```
dagguh, line 17: Oooh, I get it. We can have tests that work locally and not on CI.
Not having to provision Jira every time is gonna be faster for sure.
OTOH the dev still has to provision manually Jira at least once, so the first loop is still slow (and tribal). We need Jira to start quickly. What's the duration of `JiraCoreFormula.provision()`?

pbruski: I have data for the full run - it's 12 minutes vs 1m30s . Obviously, if you need to test just one action, it becomes 12minutes vs 10-20 seconds...

dagguh: That's convincing!

dagguh: Do you need this for ShipIt? I'd like to review some small stuff here, but I'm participating in the ShipIt right now, so I'm deferring PRs by default

pbruski: No, keep shipping!

mzyromski-atlassian: As this test is intended for manual execution only it won't get executed on CI which makes it prone to degradation in the future. That said, the gain for a dev is so big that it indeed might make sense to have it.

mzyromski-atlassian, line 17: nit: typos ("JIra", dot before "and")

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/JiraCoreScenarioIT.kt
```diff
@@ -32,111 +15,18 @@ class JiraCoreScenarioIT {
...
-                return user
-            }
-
-            override fun remember(memories: Collection<User>) {
-                throw Exception("not implemented")
-            }
-        }
-        var firstBackupElement = true
-        var secondBackupElement = true
+        val jiraVersion = System.getenv("JIRA_SOFTWARE_VERSION") ?: "8.0.0"
```
mzyromski-atlassian, line 18: How about accessing the value from `AbstractJiraCoreScenario`? I'm afraid that having it here and there (where it's actually only used to display a log message) will lead to people making mistakes in the future when e.g. changing a default value.

pbruski: Sure!

PR #72, opened 2020-07-22
JPERF-654: Accommodate to AUI 9 where CSS class for flag close button changed

mzyromski-atlassian:


src/main/kotlin/com/atlassian/performance/tools/jiraactions/page/NotificationPopUps.kt
```diff
@@ -10,7 +10,7 @@ import java.time.Duration
 
 
 internal class NotificationPopUps(private val driver: WebDriver) {
-    private val auiFlagCloseLocator = By.cssSelector(".aui-flag .icon-close")
+    private val auiFlagCloseLocator = By.cssSelector(".aui-flag .icon-close, .aui-flag .aui-close-button")
```
wyrzyk, line 13: Does it work for pre-AUI 9 Jira versions?

mzyromski-atlassian: Yup. We have ITs on 7.x and 8.0. Plus, I verified it manually on `8.12-SNAPSHOT` (with AUI 9) and `8.5.x (pre AUI 9).

dagguh: In CSS, the comma gives an alternative selector. So anything that matched before will still match. This only adds support to AUI 9 without removing the old one.

PR #77, opened 2020-10-29
JPERF-681: Make Create Issue action compatible with recent changes to create issue dialog in Jira

mzyromski-atlassian:


src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssueCreateDialog.kt
```diff
@@ -72,24 +72,32 @@ class IssueCreateDialog(
...
+            driver.wait(not(elementToBeClickable(newAllFieldsLinkLocator)))
         }
+
         dismissConfigureFieldsDialog()
         return this
     }
 
+    private fun getVisibleAllFieldsLinkLocator(dialogClass: String) : By {
+        val sectionIndex = if (driver.findElement(By.cssSelector(".$dialogClass dl:nth-of-type(1)")).isDisplayed) 1 else 2
+        return By.xpath("//div[@class='$dialogClass']//dl[$sectionIndex]//dd[1]//a")
```
dagguh, line 98: Ah, to dream about semantic HTML 😉 
Perhaps the frontend team could make the markup more robot-friendly? I asked 🤞

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssueCreateDialog.kt
```diff
@@ -72,24 +72,32 @@ class IssueCreateDialog(
...
-        val locator = By.xpath("//div[@id='$configureFieldsDialogId']//dd[1]//a")
-        if (driver.isElementPresent(locator)) {
-            driver.wait(elementToBeClickable(locator)).click()
-            driver.wait(invisibilityOfElementLocated(dialogLocator))
-            driver.wait(visibilityOfElementLocated(dialogLocator))
+
+        val allFieldsLinkLocator = getVisibleAllFieldsLinkLocator(configureFieldsDialogClass)
+        if (driver.isElementPresent(allFieldsLinkLocator)) {
+            driver.wait(elementToBeClickable(allFieldsLinkLocator)).click()
+            val newAllFieldsLinkLocator = getVisibleAllFieldsLinkLocator(configureFieldsDialogClass)
```
dagguh, line 88: Oh snap, we gotta re-find it here? Was the `jira-actions` dev loop sufficient to figure this out?

dagguh (approved):

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssueCreateDialog.kt
mzyromski-atlassian: Nope, this is a drive-by fix actually. Turned out that this code, that switches from "Custom" to "All" failed, most likely due to race condition. I discovered it during local testing.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssueCreateDialog.kt
mzyromski-atlassian: Yup. :) Another option would be to have shared page objects, but that sound too crazy.

dagguh: When we started JPT, we wanted to reuse the `atlassian-jira-pageobjects` (or something), but it came with a few problems:
* 300+ MB of dependencies to put into the shadow JAR
* Maven POM magic to autoinclude a Chrome/Firefox executable based on your current OS
    * they were out of date fast - e.g. Chome 30
    * if you run JPT on Windows, you don't want the `googlechrome.exe` on the VU node which runs Ubuntu
* despite honest effort, we couldn't figure out how to actually run this code - too many magical annotations and reflection
* very restrictive patterns - [page objects are a flexible pattern](https://github.com/SeleniumHQ/selenium/wiki/PageObjects)
* not sure if they are publicly distributed

PR #81, opened 2021-04-16
CHER-277: Fixing getIssueKeys to work for Cloud environment for WorkOnSearch key experience when JPT runs tests

jovdjune:
Taking the following approach into action:
"Add an if/else statement in your code so that getIssueKeys() checks whether issue-link-key elements exist in the page and if not it checks if issue-link elements exist (which is the new name for those elements in cloud"
from the #help-jpt channel as we got an approval to create a PR for this work.

### DIFF
--- a/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssueNavigatorPage.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssueNavigatorPage.kt
@@ -38,12 +38,20 @@ class IssueNavigatorPage(
         return this
     }
 
-    fun getIssueKeys(): Set<String> {
+    private fun getIssueKeys(cssSelector: String): Set<String> {
         val issueKeys: List<String> = JavaScriptUtils.executeScript(driver,
-            "return Array.from(document.getElementsByClassName('issue-link-key'), i => i.innerText.trim())"
+            "return Array.from(document.getElementsByClassName('${cssSelector}'), i => i.innerText.trim())"
         )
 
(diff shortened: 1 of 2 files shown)


mzyromski-atlassian:
For some reason, GH didn't pick up TravisCI build. @jovdjune, do you have the project configured on your TravisCI account?

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssueNavigatorPage.kt
```diff
@@ -38,14 +38,22 @@ class IssueNavigatorPage(
...
         )
 
         return HashSet(issueKeys)
     }
 
+    fun getIssueKeys(): Set<String> {
+        return getIssueKeys("issue-link-key")
+    }
+
+    fun getIssueKeysCloud(): Set<String> {
```
dagguh, line 53: Can Cloud-specific behavior be kept in the [inlined CloudIssueNavigator](https://github.com/atlassian/jces-1209/blob/b036176c7c568f0e1f8e75e265c9008244c1a1a6/custom-vu/src/main/kotlin/jces1209/vu/page/issuenavigator/CloudIssueNavigator.kt)?

jovdjune: @mzyromski-atlassian I don't have a TravisCI account, will the lack of this block the potential merge?

dagguh:
The new `IssueNavResultsView` interface allows customization of `listIssueKeys`

PR #84, opened 2021-04-22
Introduce GitHub Actions configuration

mzyromski-atlassian:
GHA CI will be initially red as we need to bump testcontainers library in order for the tests to be green. I'll do it in a separate PR, so that we can see the actual results on GHA.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,42 @@
+name: CI
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+  workflow_dispatch:
+    inputs:
+      release:
+        description: 'Release? yes/no'
+        default: 'no'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+    - name: Cache Gradle
+      uses: actions/cache@v2
+      with:
+        path: ~/.gradle
+        key: ${{ runner.os }}-${{ hashFiles('gradle') }}
+    - name: Build
+      run: ./gradlew build
+    - name: Upload test reports
+      if: always()
+      uses: actions/upload-artifact@v2
+      with:
+        name: test-reports
+        path: build/reports/tests
+    - name: Release
+      if: github.event.inputs.release == 'yes'
+      env:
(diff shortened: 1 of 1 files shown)


mzyromski-atlassian:
Also, I'll temporarily remove Travis CI check for the repo so that I can merge this and subsequent GHA-related PRs. I'll introduce GHA check after all the work is done.

mzyromski-atlassian:
Additional notes:
- We're migrating to GHA because we have issues with Travis CI
- Currently, on Travis, we have CI configured to run several jobs that test the library against different Jira versions. The proposed GHA configuration is simpler and will execute tests on a single, default version only (`8.0.0`). I'll create a separate PR to improve it.

dagguh (approved):
https://github.com/mzyromski-atlassian/jira-actions/actions/runs/774072197
Looks good and it's green 👍

.github/workflows/ci.yml
```diff
@@ -0,0 +1,42 @@
...
+      with:
+        name: test-reports
+        path: build/reports/tests
+    - name: Release
+      if: github.event.inputs.release == 'yes'
+      env:
+        atlassian_private_username: ${{ secrets.ARTIFACTORY_USERNAME }}
+        atlassian_private_password: ${{ secrets.ARTIFACTORY_API_KEY }}
+      run: |
+        ./gradlew release -Prelease.customUsername=${{ secrets.REPOSITORY_ACCESS_TOKEN }}
```
NeveHanter, line 41: Can't you provide that through environment variable, some file or stdin? Also shouldn't that be `"` enclosed?

mzyromski-atlassian: > Can't you provide that through environment variable, some file or stdin? Also shouldn't that be `"` enclosed?

What do you refer to, `secrets.*` variables? Those are defined in GH -- some on per-project and some on per-organisation basis.

NeveHanter: Yes, you can ignore the thing about providing the secret, I've found out that GH redact these in their plain format.

About the enclosing, there's a note in the documentation that suggests enclosing secrets:
https://docs.github.com/en/actions/reference/encrypted-secrets#using-encrypted-secrets-in-a-workflow
> If you must pass secrets within a command line, then enclose them within the proper quoting rules. Secrets often contain special characters that may unintentionally affect your shell. To escape these special characters, use quoting with your environment variables. For example:

PR #87, opened 2021-04-26
CHER-277: making getIssueKeys method open so it can be overridden and adding a getDriver method as it's currently invisible in the super class

jovdjune:


mzyromski-atlassian (approved):

mzyromski-atlassian:
Released in `3.15.0`.

PR #93, opened 2021-07-01
MNSTR-5245: Measure App Header render time

mciesielski2:
I've added the logic for getting the element performance, extracted pieces of the performance metrics logic from `JavascriptW3cPerformanceTimeline` into separate files and made gathering each metric opt-out-able without (hopefully) breaking the API backwards compatibility.

Draft PRs in virtual-users that demonstrates how does it all fit together:
https://github.com/atlassian/virtual-users/pull/58/files#r662501980

The proof this one works for the regression test is in the attached file (action metrics of a virtual user, check for the "elements" array):
[action-metrics.jpt.txt](https://github.com/atlassian/jira-actions/files/6750003/action-metrics.jpt.txt)


mciesielski2:
:thinking: 
Now that I think of it - the name does not make sense. These changes are generic and will measure all the ElementTimingAPI events. I'll rename it and branches/commits soon.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/w3c/harvesters/JsElements.kt
```diff
@@ -0,0 +1,43 @@
...
+import org.openqa.selenium.TimeoutException
+
+
+internal fun getJsElementsPerformance(javascript: JavascriptExecutor): List<PerformanceElementTiming> {
+    return try {
+        val jsElements =
+            javascript.executeAsyncScript("new PerformanceObserver(list => arguments[0](list.getEntries())).observe({type: 'element', buffered: true})")
+        parseElements(jsElements)
+    } catch (timeoutEx: TimeoutException) {
+        // No element performance entries on page.
```
dagguh, line 14: Is there a way to detect lack of element timings without waiting for a timeout?

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/w3c/RecordedPerformanceEntries.kt
```diff
@@ -6,5 +6,6 @@ package com.atlassian.performance.tools.jiraactions.api.w3c
  */
 class RecordedPerformanceEntries internal constructor(
     val navigations: List<PerformanceNavigationTiming>,
-    val resources: List<PerformanceResourceTiming>
+    val resources: List<PerformanceResourceTiming>,
+    val elements: List<PerformanceElementTiming>?
```
dagguh, line 10: So far, if navigations or resources were missing, they were an empty list.
Let's do the same for elements for consistency.
If we find that nullable lists are a better pattern, we should do it for all the fields in another PR.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/w3c/JavascriptW3cPerformanceTimeline.kt
```diff
@@ -1,117 +1,31 @@
...
  * Obtains entries from [javascript].
  */
 class JavascriptW3cPerformanceTimeline(
-    private val javascript: JavascriptExecutor
+    javascript: JavascriptExecutor,
+    withNavigationPerformance: Boolean = true,
+    withResourcesPerformance: Boolean = true,
+    withElementPerformance: Boolean = true
 ) : W3cPerformanceTimeline {
+    private var getNavigationPerformance: () -> List<PerformanceNavigationTiming> =
```
dagguh, line 17: `Supplier` fileds are basically functions:
```suggestion
    private fun getNavigationPerformance() : List<PerformanceNavigationTiming> =
```

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/w3c/JavascriptW3cPerformanceTimeline.kt
```diff
@@ -1,117 +1,31 @@
...
 import org.openqa.selenium.JavascriptExecutor
-import java.time.Duration
 
 /**
  * Obtains entries from [javascript].
  */
 class JavascriptW3cPerformanceTimeline(
-    private val javascript: JavascriptExecutor
+    javascript: JavascriptExecutor,
+    withNavigationPerformance: Boolean = true,
```
dagguh, line 13: I was thinking about decorators, but a good old `Boolean` seems to be a way better fit. Nice one!

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/w3c/JavascriptW3cPerformanceTimeline.kt
```diff
@@ -1,117 +1,31 @@
...
+    withResourcesPerformance: Boolean = true,
+    withElementPerformance: Boolean = true
```
dagguh, line 15: Kotlin default arguments are a trap. They use compilation tricks, which leave messy API, which cannot be cleanly used from Java.
We need to replace this constructor with a builder. A builder is a good vessel for default values.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/w3c/PerformanceElementTiming.kt
```diff
@@ -0,0 +1,17 @@
+package com.atlassian.performance.tools.jiraactions.api.w3c
+
+import java.time.Duration
+
+
+/**
+ * Represents the [PerformanceElementTiming](https://wicg.github.io/element-timing/#performanceelementtiming).
```
dagguh, line 7: Thanks for the reference 👌

src/main/kotlin/com/atlassian/performance/tools/jiraactions/w3c/VerboseJsonFormat.kt
```diff
@@ -26,7 +27,10 @@ internal class VerboseJsonFormat {
                 .map { deserializeNavigationTiming(it) },
             resources = getJsonArray("resources")
                 .map { it.asJsonObject() }
-                .map { deserializeResourceTiming(it) }
+                .map { deserializeResourceTiming(it) },
+            elements = getJsonArray("elements")
+                ?.map { it.asJsonObject() }
+                ?.map { deserializeElementTiming(it) }
```
dagguh, line 33: Indeed, `getJsonArray("elements")` could be `null`, therefore the null-safe dereference is needed: `?.`
That's because of old JSON files, which should remain parseable for compatibility.

But if we set the `elements` property to be non-null for consistency, then we need to default to an empty list (via the Elvis operator: `?:`):
```suggestion
            elements = getJsonArray("elements")
                ?.map { it.asJsonObject() }
                ?.map { deserializeElementTiming(it) }
                ?: emptyList()
```

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/w3c/harvesters/JsElements.kt
dagguh: Note that `RecordedPerformanceEntries` (including its `resources` field) is taken as a snapshot at the end of a JPT `Action`.
Recording of the performance entries does not wait for anything to load anymore. The JPT Action did the waiting already.

- [ ] avoid waiting in `JavascriptW3cPerformanceTimeline.record`

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/w3c/harvesters/JsElements.kt
```diff
@@ -0,0 +1,43 @@
...
+
+import com.atlassian.performance.tools.jiraactions.api.w3c.PerformanceElementTiming
+import org.openqa.selenium.JavascriptExecutor
+import org.openqa.selenium.TimeoutException
+
+
+internal fun getJsElementsPerformance(javascript: JavascriptExecutor): List<PerformanceElementTiming> {
+    return try {
+        val jsElements =
+            javascript.executeAsyncScript("new PerformanceObserver(list => arguments[0](list.getEntries())).observe({type: 'element', buffered: true})")
```
dagguh, line 11: What does the `buffered: true` do?

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/w3c/PerformanceElementTiming.kt
```diff
@@ -0,0 +1,17 @@
+package com.atlassian.performance.tools.jiraactions.api.w3c
```
dagguh, line 1: Let's add the new API to the CHANGELOG.md

dagguh (changes requested):
A couple details need some polishing, but overall it looks really good!

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/w3c/harvesters/Common.kt
```diff
@@ -0,0 +1,11 @@
+package com.atlassian.performance.tools.jiraactions.api.w3c.harvesters
```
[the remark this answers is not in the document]
mciesielski2: [ ] Clean this up.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/w3c/harvesters/JsElements.kt
rzymek01: You can attach JS listener after some time and still get measures that were registered/emitted by the browser before loading JS.

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/AbstractJiraCoreScenario.kt
```diff
@@ -26,7 +30,8 @@ abstract class AbstractJiraCoreScenario {
         val metrics = mutableListOf<ActionMetric>()
         val actionMeter = ActionMeter.Builder(
             output = CollectionActionMetricOutput(metrics)
-        ).build()
+        ).appendPostMetricHook(DrillDownHook(JavascriptW3cPerformanceTimeline(driver as JavascriptExecutor)))
```
dagguh, line 33: The opt-in is roundabout, isn't it? That's because historically we introduced a *specific* API for drilldown (`W3cPerformanceTimeline`) and only later we *generalized* it to `PostMetricHook`. Now the `DrillDownHook` acts as a bridge.

We can make it simpler if we make `JavascriptW3cPerformanceTimeline` implement the `PostMetricHook` directly.
We couldn't do that with `W3cPerformanceTimeline` due to SPI compatibility, but we can do that with the impl, because it's `final`. Then we could do this:
```suggestion
        ).appendPostMetricHook(JavascriptW3cPerformanceTimeline(driver as JavascriptExecutor))
```

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/w3c/harvesters/JsElements.kt
mciesielski2: Yep, seems like this is possible with the `.takeRecords()` method ([doc here](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceObserver/takeRecords)).
Testing in progress...

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/w3c/RecordedPerformanceEntries.kt
mciesielski2: I did that because I thought it would not be a breaking change. Is it OK to just add a new field here from the API perspective?

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/w3c/JavascriptW3cPerformanceTimeline.kt
mciesielski2: Makes sense!
- [ ] Do it!

src/main/kotlin/com/atlassian/performance/tools/jiraactions/w3c/VerboseJsonFormat.kt
mciesielski2: - [ ] Do it!

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/w3c/PerformanceElementTiming.kt
mciesielski2: - [] Do it!

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/AbstractJiraCoreScenario.kt
mciesielski2: Makes sense :+1:  :D

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/w3c/JavascriptW3cPerformanceTimeline.kt
mciesielski2: Seems like it won't work - the function implementation differs based on the `with...` field. I think it has to remain a var :/ 
Or it's just me not being a Kotlin master yet ;P

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/w3c/RecordedPerformanceEntries.kt
dagguh: Yes, because the constructor is internal

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/w3c/JavascriptW3cPerformanceTimeline.kt
dagguh: It's because these are wrapped in too many `{ }`, turning them into lambdas.
Here's a few adjustments: https://github.com/atlassian-forks/jira-actions/pull/1

src/main/kotlin/com/atlassian/performance/tools/jiraactions/w3c/VerboseJsonFormat.kt
```diff
@@ -14,6 +14,7 @@ internal class VerboseJsonFormat {
         Json.createObjectBuilder()
             .add("navigations", navigations.map { serializeNavigationTiming(it) }.toJsonArray())
             .add("resources", resources.map { serializeResourceTiming(it) }.toJsonArray())
+            .add("elements", (elements?.map { serializeElementTiming(it) } ?: emptyList()).toJsonArray())
```
dagguh, line 17: `elements` are no longer nullable, so this can become consistent too:
```suggestion
            .add("elements", elements.map { serializeElementTiming(it) }.toJsonArray())
```
IntelliJ is tracking the nullability in Kotlin correctly and even warns about an unnecessary safe call here

dagguh (approved):

dagguh:
Looks good. Only nice-to-haves are remaining:

- [ ] builder API adjustments
- [ ] rebase/squash commits
- [ ] [JPERF](https://ecosystem.atlassian.net/browse/JPERF) ticket + changelog link

These tasks are on best-effort basis and do not block the merge.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/w3c/JavascriptW3cPerformanceTimeline.kt
mciesielski2: Thanks for that!

dagguh (approved):

dagguh (approved):

src/main/kotlin/com/atlassian/performance/tools/jiraactions/w3c/VerboseJsonFormat.kt
mciesielski2: - [ ] Do it

mciesielski2:
Damn, I tried to rename the branch and that closed the PR.
I'm opening a new one, tidied up this time.
Sorry, not used to github!

dagguh:
No worries 😉 
PS. Branch names don't matter, because they get deleted anyway. We ban merge commits, so there's no trace of the branches in the Git.

PR #96, opened 2021-07-27
Adjusting the Configure Fields dialog on issue view to recent accessi…

Nubzor:
Since last accessibility changes on create issue view the performance tests started failing as the UI has changed. 

I've added a new method(s) in order to keep backward compatibility and started using the ajs-build-number meta tag as version differentiator not sure if this is a correct approach tho

I ran local tests to make sure that both paths are working correctly.

### DIFF
--- a/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssueCreateDialog.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssueCreateDialog.kt
@@ -7,6 +7,7 @@ import com.atlassian.performance.tools.jiraactions.page.SingleSelect
 import org.openqa.selenium.*
 import org.openqa.selenium.support.ui.ExpectedCondition
 import org.openqa.selenium.support.ui.ExpectedConditions.*
+import org.openqa.selenium.support.ui.Select
 import java.time.Duration
 import java.util.function.Supplier
 
@@ -70,10 +71,13 @@ class IssueCreateDialog(
      * Click 'Configure Fields' to display 'All' to ensure
      * all mandatory fields are displayed in creation dialog.
      *
+     * The view has changed since Jira 8.19 so differentiator is needed
      */
     fun showAllFields(): IssueCreateDialog {
         val configureFieldsDialogClass = "qf-picker"
         val dialogLocator = By.className(configureFieldsDialogClass)
+        val selectLocator = getFieldsSelectLocator(configureFieldsDialogClass);
+        val newConfigureFieldsUi = driver.isElementPresent(selectLocator)
 
         try {
             openConfigureFieldsDialog(dialogLocator)
@@ -82,11 +86,23 @@ class IssueCreateDialog(
             openConfigureFieldsDialog(dialogLocator)
         }
 
-        val allFieldsLinkLocator = getVisibleAllFieldsLinkLocator(configureFieldsDialogClass)
-        if (driver.isElementPresent(allFieldsLinkLocator)) {
-            driver.wait(elementToBeClickable(allFieldsLinkLocator)).click()
-            val newAllFieldsLinkLocator = getVisibleAllFieldsLinkLocator(configureFieldsDialogClass)
-            driver.wait(not(elementToBeClickable(newAllFieldsLinkLocator)))
+        if (newConfigureFieldsUi) {
+            driver
+                .findElements(selectLocator)
+                .singleOrNull()
+                ?.let { Select(it) }
+                ?.takeIf { it.firstSelectedOption.text != "All Fields" }
+                ?.let {
+                    it.selectByVisibleText("All Fields")
+                    driver.wait(elementToBeClickable(selectLocator))
+                }
+        } else {
+            val allFieldsLinkLocator = getVisibleAllFieldsLinkLocator(configureFieldsDialogClass)
+            if (driver.isElementPresent(allFieldsLinkLocator)) {
+                driver.wait(elementToBeClickable(allFieldsLinkLocator)).click()
+                val newAllFieldsLinkLocator = getVisibleAllFieldsLinkLocator(configureFieldsDialogClass)
+                driver.wait(not(elementToBeClickable(newAllFieldsLinkLocator)))
+            }
         }
 
         dismissConfigureFieldsDialog()
@@ -96,6 +112,10 @@ class IssueCreateDialog(
     private fun getVisibleAllFieldsLinkLocator(dialogClass: String) : By {
         val sectionIndex = if (driver.findElement(By.cssSelector(".$dialogClass dl:nth-of-type(1)")).isDisplayed) 1 else 2
         return By.xpath("//div[@class='$dialogClass']//dl[$sectionIndex]//dd[1]//a")
+    }
+
+    private fun getFieldsSelectLocator(dialogClass: String): By {
+        return By.cssSelector(".$dialogClass #configure-fields");
     }
 
     private fun openConfigureFieldsDialog(popupLocator: By) {


mzyromski-atlassian (approved):

mzyromski-atlassian:
I'll release new `jira-actions` version after this PR is merged.

mzyromski-atlassian (dismissed):

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssueCreateDialog.kt
```diff
@@ -82,22 +84,46 @@ class IssueCreateDialog(
...
         }
 
         dismissConfigureFieldsDialog()
         return this
     }
 
+    private fun isNewConfigureFieldsUiAvailable(): Boolean {
+        return driver.findElement(By.cssSelector("meta[name=ajs-build-number]"))
+            .getAttribute("content")
+            .toInt() >= 819000
```
dagguh, line 115: Does the new AJS provide semantic DOM to rely on?

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssueCreateDialog.kt
```diff
@@ -82,22 +84,46 @@ class IssueCreateDialog(
...
 
-        val allFieldsLinkLocator = getVisibleAllFieldsLinkLocator(configureFieldsDialogClass)
-        if (driver.isElementPresent(allFieldsLinkLocator)) {
-            driver.wait(elementToBeClickable(allFieldsLinkLocator)).click()
-            val newAllFieldsLinkLocator = getVisibleAllFieldsLinkLocator(configureFieldsDialogClass)
-            driver.wait(not(elementToBeClickable(newAllFieldsLinkLocator)))
+        if (newConfigureFieldsUi) {
+            val selectLocator = getFieldsSelectLocator(configureFieldsDialogClass);
+
+            if (driver.isElementPresent(selectLocator)) {
```
dagguh, line 90: Can this condition become the differentiator?

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssueCreateDialog.kt
```diff
@@ -82,22 +84,46 @@ class IssueCreateDialog(
...
+                val s = Select(driver.findElement(selectLocator))
```
dagguh, line 91: We can squash this `findElement` call and `isElement` (calling `findElements` underneath) into a single WD call.

Classic style:
```kotlin
val select = driver
    .findElements(selectLocator)
    .singleOrNull()
    ?.let { Select(it) }
if (select != null && select.firstSelectedOption.text != "All Fields") {
    select.selectByVisibleText("All Fields")
    driver.wait(elementToBeClickable(selectLocator));
}
```

Stream style:
```kotlin
driver
    .findElements(selectLocator)
    .singleOrNull()
    ?.let { Select(it) }
    ?.takeIf { it.firstSelectedOption.text != "All Fields" }
    ?.let {
        it.selectByVisibleText("All Fields")
        driver.wait(elementToBeClickable(selectLocator));
    }
```

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssueCreateDialog.kt
```diff
@@ -70,10 +70,12 @@ class IssueCreateDialog(
      * Click 'Configure Fields' to display 'All' to ensure
      * all mandatory fields are displayed in creation dialog.
      *
+     * The view has changed since Jira 8.19 so differentiator is needed
```
dagguh, line 73: Would the new [semantic DOM attributes from 8.14](https://bulldog.internal.atlassian.com/browse/MNSTR-4635) cover this?

dagguh, line 73: If you want to increase CI coverage, the Jira versions are set in `.github/workflows/ci.yml`

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssueCreateDialog.kt
```diff
@@ -44,7 +45,6 @@ class IssueCreateDialog(
 
     fun selectProject(projectName: String) = form.waitForRefresh(Supplier {
         projectField.select(projectName)
-        waitUntilSummaryIsFocused()
```
dagguh, line 47: Huh, this is deleted, but tests didn't go red. Maybe this waiting wasn't necessary, but let's change that in a separate PR.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssueCreateDialog.kt
Nubzor: What do you mean by a new AJS? Meta tags are rendered within <head> and are in Jira since... ages?

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssueCreateDialog.kt
Nubzor: This is a good question, if this is true we could get rid of Jira's version comparison.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssueCreateDialog.kt
Nubzor: Will do, much clearer.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssueCreateDialog.kt
Nubzor: It should since currently we are testing up to 8.13 (based on the ci.yml)

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssueCreateDialog.kt
Nubzor: It started to fail on this lane after the accessibility changes tho when I removed it the previous versions seems to work just fine. 

I'll follow your advice and prepare a separate PR.

wyrzyk: Often tests can go green without waits. It could just become flaky.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssueCreateDialog.kt
dagguh: Sorry, that's my Jira frontend ignorance speaking :sweat_smile: 
I meant: does Jira 8.19 provide semantic (long-term) selectors for new GUI? it's a practice similar to feature detection vs browser detection

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssueCreateDialog.kt
dagguh: And with smaller WD overhead :slightly_smiling_face:

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssueCreateDialog.kt
dagguh: The attributes linked above are a response to https://github.com/atlassian/jira-actions/pull/77
The "then new" GUI was detected via a very complicated and very unstable selector in `fun getVisibleAllFieldsLinkLocator`. Now we're adding "new new GUI". Just like  `New Folder (6)` on desktop or `essay-final3-LAST (2).doc` :wink:  There's always gonna be something newer, unless we future-proof it.

The future-proofing idea is: get a stable (covered by HTML tests :heavy_check_mark:) semantic (user terms, not tech :heavy_check_mark:) DOM and rely on that for the long term. These were shipped in 8.14. Semantic attributes should be detected first, then fall back to legacy non-semantic code (like `dl:nth-of-type(1)` and `/div[@class='$dialogClass']//dl[$sectionIndex]//dd[1]//a`).

The `jira-actions` didn't consume the attributes yet, so we're not sure if they work at all and if they'll work with 8.19.
Unfortunately, it's hard to test unreleased Jira versions in this module here, because these tests don't reuse the  `infrastructure` API, which would allow that. We'll get that fixed after I ship [JPERF-273](https://ecosystem.atlassian.net/browse/JPERF-273) in innovation time.

pczuj: > it's hard to test unreleased Jira versions in this module here

First EAP of 8.19 should be available soon (see Atlassian internal Slack on #jira-server-release).

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssueCreateDialog.kt
pczuj: Won't such fallbacks/decisions impact the measure? Shouldn't the decision be done in the action inself, before we start to measure? i.e. VU knows that we are on version X, so it uses page objects compatible with X

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssueCreateDialog.kt
Nubzor: For the Configure Fields it does - there is a unique identifier and some data-test-id attributes, anyway, as per another comment I'll base on feature detection and on the presence of the new element

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssueCreateDialog.kt
Nubzor: Currently, I don't as 8.19 is not released yet.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssueCreateDialog.kt
Nubzor: Probably the measure will be affected a little bit tho it will be affected by the UI change itself. 

Also, I am not sure if I'll be able to access WebDriver object in the action itself. 

I am open to suggestions.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssueCreateDialog.kt
Nubzor: Thanks for clarifying, I definitely agree.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssueCreateDialog.kt
dagguh: I don't worry about the impact here, because:
* soon, most tests will short-circuit on the new GUI - the natural bias towards testing new versions will deepen with time
* a couple of WD calls is not a big cost - it's not an n+1
* if it distinguishes visually or structurally differences, then the human user also needs to spend time for recognition
    * the impact on VU might be too small actually
    * effectively, VUs using this code represent humans who are more familiar with new GUI and less familiar with old GUI
* we might drop support for old Jiras at some point, we can remove the fallbacks then

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssueCreateDialog.kt
dagguh: The other thread is about 8.14 stuff. If 8.14 selectors are sufficient for 8.19, then this convo can be resolved.

dagguh: So we got 8.19 compat without version detection, noice!

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssueCreateDialog.kt
dagguh: So 8.14 future proofing isn't guaranteed to work straight away, so we can do that in a separate PR. This PR adds value of 8.19+ support and doesn't incur much tech debt. Imma approving :smiley:

dagguh (approved):

PR #97, opened 2021-07-28
Making sure the summary field is being focused after project selection

Nubzor:
The summary field is no longer being focused after recent accessibility changes we made in Jira 8.19. 
This PR removes the wait function as the field is not going to be focused. 

Backward compatibility is going to be present as the `fill` method invoked in `CreateIssueAction.kt` (defined in `IssueCreateDialog.kt`) relies on `sendKeysWhenClickable` which clicks on the element before sending keys, the click should focus the element out of the box. 

The wait function could be added in order to prevent race conditions but this should be eventually caught by the tests.
I am mentioning Przemek (@pbruski) as he introduced the change in the first place.

### DIFF
--- a/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssueCreateDialog.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssueCreateDialog.kt
@@ -152,9 +152,10 @@ class IssueCreateDialog(
     private fun elementIsFocused(locator: By): ExpectedCondition<WebElement?> {
         return object : ExpectedCondition<WebElement?> {
             override fun apply(driver: WebDriver?): WebElement? {
-                val summary = driver!!.findElement(locator)
-
-                return if (summary == driver.switchTo().activeElement()) summary else null
+                return elementToBeClickable(locator)
+                    .apply(driver)
+                    ?.also { it.click() }
+                    ?.takeIf { it == driver!!.switchTo().activeElement() }
             }
 
             override fun toString(): String {


pbruski:
The problem with the code before my change (i.e. before I added that wait) is that it was flaky at stress load. I'd recommend detecting DOM from 8.19 instead of plain removal.
I think removal of that wait will generally work, but it will still fail at stress load (or higher load in general) and will be problematic during regression testing against older releases.

pbruski:
> The wait function could be added in order to prevent race conditions but this should be eventually caught by the tests.

Not sure if I understand you correctly, but we should not be detecting functional issues in performance tests, they are problematic to take care of even without that.

Nubzor:
Przemek, thank you for your insights, I decided to change the approach tho to avoid DOM detecting I'm using VU to focus the summary field to make it compatible across multiple Jira's versions. Once we stop supporting Jira's < 8.14 we can get rid of the code.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssueCreateDialog.kt
```diff
@@ -152,9 +152,11 @@ class IssueCreateDialog(
     private fun elementIsFocused(locator: By): ExpectedCondition<WebElement?> {
         return object : ExpectedCondition<WebElement?> {
             override fun apply(driver: WebDriver?): WebElement? {
-                val summary = driver!!.findElement(locator)
+                val element = driver!!.findElement(locator)
 
-                return if (summary == driver.switchTo().activeElement()) summary else null
+                driver.wait(elementToBeClickable(locator)).click()
+
+                return if (element == driver.switchTo().activeElement()) element else null
```
dagguh, line 159: If we don't want to wait in condition (within another wait), then we can delegate to `elementToBeClickable.apply` instead:
```kotlin
override fun apply(driver: WebDriver?): WebElement? {
    return elementToBeClickable(locator)
        .apply(driver)
        ?.also { it.click() }
        ?.takeIf { it == driver!!.switchTo().activeElement() }
}
```

mzyromski-atlassian (approved):

mzyromski-atlassian:
Merged exceptionally with red CI failing due to infra issues.

PR #103, opened 2022-04-27
Issue/mnstr 6059 implement new actions for activity tabs

ewefie:
This PR introduces two new actions for Jira: VIEW_FOCUSED_COMMENT and VIEW_HISTORY_TAB.

Because of significant differences between Tab Panel in Jira 8 and Jira 9 I had to implements two paths for those actions.
 - Focused comment - I added a new memory for comments, and to populate it I added exploring comments as a part of VIEW_ISSUE action.
The exploring part for Jira 9 is a bit more complex than in Jira 8 as the button to load more at the beginning allows only to load 10 more comments, and later, after first click, ability to load all the rest of comments is unlocked. In Jira 8 one click on the buton loads all comments at once.
- History tab - here we only wait for all entries to be loaded, but on Jira 8 all entries appear instantly, on Jira 9 only 10 entries are available right after switching to this tab, the rest has to be fetched by clicking on the button.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,14 @@ Dropping a requirement of a major version of a dependency is a new contract.
 
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/jira-actions/compare/release-3.16.3...master
+### Added
+- Add new action View History Tab
+- Add new action View Comment
+- Add new page objects for Comment Tab Panel and History Tab Panel
+- Add `ViewIssueAction.Builder`
+- Extend `ViewIssueAction` with comment remembering capability
+- Add `CommentMemory`
+- Add new method to `WebJira` that allows visiting a comment
 
 ## [3.16.3] - 2022-04-08
 [3.16.3]: https://github.com/atlassian/jira-actions/compare/release-3.16.2...release-3.16.3
--- a/README.md
+++ b/README.md
@@ -19,6 +19,8 @@ Edit Issue Submit   | While in Edit Issue Action: Press Submit → Wait for the 
 Add Comment         | Go to Add Comment for an issue form → Wait for form to load → Enter comment text → Press Submit → Wait for the summary on view issue page to show
 Add Comment Submit  | While in Add Comment Action: Press Submit → Wait for the summary on view issue page to show
 Browse Projects     | Go to Browse Projects Page → Wait for the project list to load
+View Comment        | Go to Comment Tab Panel located on View Issue Page → Wait until comment is focused
+View History Tab    | Go to History Tab Panel located on View Issue Page → Wait for all change history entries to load
 
 ## Reporting issues
 
--- a/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/ActionType.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/ActionType.kt
@@ -49,3 +49,5 @@ data class ActionType<out T>(
 @JvmField val VIEW_BOARD = ActionType("View Board") { IssuesOnBoard(it) }
 @JvmField val LOG_IN = ActionType("Log In") { Unit }
 @JvmField val SET_UP = ActionType("Set Up") { Unit }
+@JvmField val VIEW_COMMENT = ActionType("View Comment") { Unit }
+@JvmField val VIEW_HISTORY_TAB = ActionType("View History Tab") { IssueObservation(it) }
--- a/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/WebJira.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/WebJira.kt
@@ -22,7 +22,7 @@ data class WebJira(
         return RichTextEditorConfiguration(driver, accessAdmin())
     }
 
-    fun configureBackupPolicy() : BackupConfiguration {
+    fun configureBackupPolicy(): BackupConfiguration {
         navigateTo("secure/admin/ViewServices!default.jspa")
         return BackupConfiguration(driver, accessAdmin())
     }
@@ -92,6 +92,13 @@ data class WebJira(
 
     fun getJiraNode(): String {
         return driver.findElement(By.id("footer-build-information")).text
+    }
+
+    fun goToComment(
+        commentUrl: String
+    ): CommentTabPanel {
+        driver.navigate().to(commentUrl)
+        return CommentTabPanel(driver)
     }
 
     fun navigateTo(path: String) {
--- a/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/action/ViewCommentAction.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/action/ViewCommentAction.kt
@@ -1,0 +1,36 @@
+package com.atlassian.performance.tools.jiraactions.api.action
+
+import com.atlassian.performance.tools.jiraactions.api.VIEW_COMMENT
+import com.atlassian.performance.tools.jiraactions.api.WebJira
+import com.atlassian.performance.tools.jiraactions.api.measure.ActionMeter
+import com.atlassian.performance.tools.jiraactions.api.memories.CommentMemory
+import org.apache.logging.log4j.LogManager
+import org.apache.logging.log4j.Logger
+import javax.json.Json
+
+class ViewCommentAction(
+    private val jira: WebJira,
+    private val meter: ActionMeter,
+    private val commentMemory: CommentMemory
+) : Action {
+    private val logger: Logger = LogManager.getLogger(this::class.java)
+
+    override fun run() {
+        val comment = commentMemory.recall()
+        if (comment == null) {
+            logger.debug("Skipping View Comment action. I have no knowledge of comments.")
+            return
+        }
+
+        meter.measure(
+            key = VIEW_COMMENT,
+            action = { jira.goToComment(comment.url).validateCommentIsFocused(comment.id) },
+            observation = { page ->
+                Json.createObjectBuilder()
+                    .add("issueKey", page.getIssueKey())
+                    .add("commentId", comment.id)
+                    .build()
+            }
+        )
+    }
+}
--- a/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/action/ViewHistoryTabAction.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/action/ViewHistoryTabAction.kt
@@ -1,0 +1,32 @@
+package com.atlassian.performance.tools.jiraactions.api.action
+
+import com.atlassian.performance.tools.jiraactions.api.VIEW_HISTORY_TAB
+import com.atlassian.performance.tools.jiraactions.api.WebJira
+import com.atlassian.performance.tools.jiraactions.api.measure.ActionMeter
+import com.atlassian.performance.tools.jiraactions.api.memories.IssueKeyMemory
+import com.atlassian.performance.tools.jiraactions.api.observation.IssueObservation
+import org.apache.logging.log4j.LogManager
+import org.apache.logging.log4j.Logger
+
+class ViewHistoryTabAction(
+    private val jira: WebJira,
+    private val meter: ActionMeter,
+    private val issueKeyMemory: IssueKeyMemory
+) : Action {
+    private val logger: Logger = LogManager.getLogger(this::class.java)
+
+    override fun run() {
+        val issueKey = issueKeyMemory.recall()
+        if (issueKey == null) {
+            logger.debug("Skipping View History Tab action. I have no knowledge of issue keys.")
+            return
+        }
+
+        val page = jira.goToIssue(issueKey).waitForSummary()
+        meter.measure(
+            key = VIEW_HISTORY_TAB,
+            action = { page.openHistoryTabPanel().loadAllHistoryEntries() },
+            observation = { IssueObservation(issueKey).serialize() }
+        )
+    }
+}
--- a/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/action/ViewIssueAction.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/action/ViewIssueAction.kt
@@ -3,25 +3,39 @@ package com.atlassian.performance.tools.jiraactions.api.action
 import com.atlassian.performance.tools.jiraactions.api.VIEW_ISSUE
 import com.atlassian.performance.tools.jiraactions.api.WebJira
 import com.atlassian.performance.tools.jiraactions.api.measure.ActionMeter
-import com.atlassian.performance.tools.jiraactions.api.memories.Issue
-import com.atlassian.performance.tools.jiraactions.api.memories.IssueKeyMemory
-import com.atlassian.performance.tools.jiraactions.api.memories.IssueMemory
-import com.atlassian.performance.tools.jiraactions.api.memories.JqlMemory
+import com.atlassian.performance.tools.jiraactions.api.memories.*
 import org.apache.logging.log4j.LogManager
 import org.apache.logging.log4j.Logger
 import javax.json.Json
 
-class ViewIssueAction(
+class ViewIssueAction private constructor(
     private val jira: WebJira,
     private val meter: ActionMeter,
-    private val issueKeyMemory: IssueKeyMemory,
-    private val issueMemory: IssueMemory,
-    private val jqlMemory: JqlMemory
+    private val issueKeyMemory: IssueKeyMemory?,
+    private val issueMemory: IssueMemory?,
+    private val jqlMemory: JqlMemory?,
+    private val commentMemory: CommentMemory?
 ) : Action {
     private val logger: Logger = LogManager.getLogger(this::class.java)
 
+    @Deprecated("Use ViewIssueAction.Builder instead.")
+    constructor(
+        jira: WebJira,
+        meter: ActionMeter,
+        issueKeyMemory: IssueKeyMemory?,
+        issueMemory: IssueMemory?,
+        jqlMemory: JqlMemory?
+    ) : this(
+        jira = jira,
+        meter = meter,
+        issueKeyMemory = issueKeyMemory,
+        issueMemory = issueMemory,
+        jqlMemory = jqlMemory,
+        commentMemory = null
+    )
+
     override fun run() {
-        val issueKey = issueKeyMemory.recall()
+        val issueKey = issueKeyMemory?.recall()
         if (issueKey == null) {
             logger.debug("Skipping View Issue action. I have no knowledge of issue keys.")
             return
@@ -29,10 +43,11 @@ class ViewIssueAction(
         val issuePage = meter.measure(
             key = VIEW_ISSUE,
             action = { jira.goToIssue(issueKey).waitForSummary() },
-            observation = { page -> Json.createObjectBuilder()
-                .add("issueKey", issueKey)
-                .add("issueId", page.getIssueId())
-                .build()
+            observation = { page ->
+                Json.createObjectBuilder()
+                    .add("issueKey", issueKey)
+                    .add("issueId", page.getIssueId())
+                    .build()
             }
         )
         val issue = Issue(
@@ -41,7 +56,38 @@ class ViewIssueAction(
             id = issuePage.getIssueId(),
             type = issuePage.getIssueType()
         )
-        issueMemory.remember(setOf(issue))
-        jqlMemory.observe(issuePage)
+        val comments = issuePage.openCommentTabPanel()
+            .showAllComments()
+            .getComments()
+
+        issueMemory?.remember(setOf(issue))
+        commentMemory?.remember(comments)
+        jqlMemory?.observe(issuePage)
+    }
+
+    class Builder(
+        private var jira: WebJira,
+        private var meter: ActionMeter
+    ) {
+        private var issueKeyMemory: IssueKeyMemory? = null
+        private var issueMemory: IssueMemory? = null
+        private var jqlMemory: JqlMemory? = null
+        private var commentMemory: CommentMemory? = null
+
+        fun webJira(jira: WebJira) = apply { this.jira = jira }
+        fun meter(meter: ActionMeter) = apply { this.meter = meter }
+        fun issueKeyMemory(issueKeyMemory: IssueKeyMemory) = apply { this.issueKeyMemory = issueKeyMemory }
+        fun issueMemory(issueMemory: IssueMemory) = apply { this.issueMemory = issueMemory }
+        fun jqlMemory(jqlMemory: JqlMemory) = apply { this.jqlMemory = jqlMemory }
+        fun commentMemory(commentMemory: CommentMemory) = apply { this.commentMemory = commentMemory }
+
+        fun build() = ViewIssueAction(
+            jira = jira,
+            meter = meter,
+            issueKeyMemory = issueKeyMemory,
+            issueMemory = issueMemory,
+            jqlMemory = jqlMemory,
+            commentMemory = commentMemory
+        )
     }
 }
--- a/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/memories/Comment.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/memories/Comment.kt
@@ -1,0 +1,24 @@
+package com.atlassian.performance.tools.jiraactions.api.memories
+
+class Comment(
+    val id: String,
+    val url: String
+) {
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (javaClass != other?.javaClass) return false
+
+        other as Comment
+
+        if (id != other.id) return false
+        if (url != other.url) return false
+
+        return true
+    }
+
+    override fun hashCode(): Int {
+        var result = id.hashCode()
+        result = 31 * result + url.hashCode()
+        return result
+    }
+}
--- a/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/memories/CommentMemory.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/memories/CommentMemory.kt
@@ -1,0 +1,3 @@
+package com.atlassian.performance.tools.jiraactions.api.memories
+
+interface CommentMemory : Memory<Comment>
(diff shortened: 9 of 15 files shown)


lipinskipawel:
We should add information to the changelog about new API's.

lipinskipawel (commented):
We should add information to the changelog about new API's.

lipinskipawel (dismissed):

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/JiraCoreScenario.kt
```diff
@@ -81,7 +92,9 @@ class JiraCoreScenario constructor() : Scenario {
             viewDashboard to 15, // note that we may end up generating more views if we need to display top nav
             editIssue to 5,
             addComment to 2,
-            browseProjects to 5
+            browseProjects to 5,
+            viewFocusedComment to 5,
+            viewHistoryTabAction to 5
```
pczuj, line 97: As menitoned in our offline discussion I'm not sure if we should extend the existing scenario. Does it affect our API somehow? If so should it be noted in CHANGELOG.md?

pczuj: At the same time we need to ensure that the action is part of some test in this repo.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/memories/Comment.kt
```diff
@@ -0,0 +1,6 @@
+package com.atlassian.performance.tools.jiraactions.api.memories
+
+data class Comment(
+    val id: String,
```
pczuj, line 4: This is probably insignificant, but should we actually remember comment ID? Are users doing that? When I'm relating to an issue comment I only use URL. The ID seems to be a Jira's implementation details.

Nevertheless we are remembering ID's already in memories for other data types, so maybe it's fine.

CHANGELOG.md
```diff
@@ -24,7 +24,13 @@ Adding a requirement of a major version of a dependency is breaking a contract.
...
 
 ## [Unreleased]
-[Unreleased]: https://github.com/atlassian/jira-actions/compare/release-3.16.1...master
+[Unreleased]: https://github.com/atlassian/jira-actions/compare/release-3.17.0...master
+
+## [3.17.0] - 2022-04-13
+[3.17.0]: https://github.com/atlassian/jira-actions/compare/release-3.16.2...release-3.17.0
+### Added
+- Add new action View Focused Comment
+- Add new action View History Tab
```
pczuj, line 33: Do we want to create issues in JPERF project for that? I don't think it's necessary, but may explain to some readers why we are adding those.

README.md
```diff
@@ -19,6 +19,8 @@ Edit Issue Submit   | While in Edit Issue Action: Press Submit → Wait for the
 Add Comment         | Go to Add Comment for an issue form → Wait for form to load → Enter comment text → Press Submit → Wait for the summary on view issue page to show
 Add Comment Submit  | While in Add Comment Action: Press Submit → Wait for the summary on view issue page to show
 Browse Projects     | Go to Browse Projects Page → Wait for the project list to load
+View Focused Comment| Go to Comment Tab Panel → Wait for focused comment to load
+View History Tab    | Go to History Tab Panel → Wait for all change history entries to load
```
pczuj, line 23: Nit: Maybe it would be good to mention that the actions are about sub-view of issue view. I'm not sure if I'd know this from the message without the context I currently have.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/ActionType.kt
```diff
@@ -49,3 +49,5 @@ data class ActionType<out T>(
 @JvmField val VIEW_BOARD = ActionType("View Board") { IssuesOnBoard(it) }
 @JvmField val LOG_IN = ActionType("Log In") { Unit }
 @JvmField val SET_UP = ActionType("Set Up") { Unit }
+@JvmField val VIEW_FOCUSED_COMMENT = ActionType("View Focused Comment") {}
```
pczuj, line 52: Nit: If we are changing the convention to not add `Unit` explicitly then we should remove it from other declarations.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/action/ViewHistoryTabAction.kt
```diff
@@ -0,0 +1,28 @@
...
+    private val logger: Logger = LogManager.getLogger(this::class.java)
+
+    override fun run() {
+        val issueKey = issueKeyMemory.recall()
+        if (issueKey == null) {
+            logger.debug("Skipping View History Tab action. I have no knowledge of issue keys.")
+            return
+        }
+
+        meter.measure(key = VIEW_HISTORY_TAB) {
```
pczuj, line 24: I think we should add issue information to the measure observation (3rd argument of `meter.measure()`). If I'm correct without it we won't be able to reason based on results if specific issues have this view slower than others.

pczuj: Same for `VIEW_FOCUSED_COMMENT`

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/WebJira.kt
```diff
@@ -94,6 +94,20 @@ data class WebJira(
         return driver.findElement(By.id("footer-build-information")).text
     }
 
+    fun goToFocusedComment(
```
pczuj, line 97: Nit: How about `goToComment`? I believe the meaning is the same and it's simpler.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/WebJira.kt
```diff
@@ -94,6 +94,20 @@ data class WebJira(
...
+        commentUrl: String
+    ): CommentTabPanel {
+        driver.navigate().to(commentUrl)
+        return CommentTabPanel(driver)
+    }
+
+    fun goToHistoryTab(
+        issueKey: String
+    ): HistoryTabPanel {
+        navigateTo("/browse/$issueKey?page=com.atlassian.jira.plugin.system.issuetabpanels:changehistory-tabpanel")
```
pczuj, line 107: Topic for discussion, possibly not as part of this PR:

Should we glue such complex URLs in the action code? I understand that users may do so for simple ones like `<Jira>/browse/<issue_key>` or `<Jira>/issues/?jql=<JQL>` (at least I sometimes do), but are people adding this `page=com.atlassian.jira.plugin.system.issuetabpanels:changehistory-tabpanel` URL suffix instead of clicking the tab?

The same goes for other existing `navigateTo()` in this class so I'm not for changing the approach in this PR.

On the other hand static URLs like `secure/Dashboard.jspa` can be bookmarked. That saying issue view with history tab can also be bookmarked, however then we should probably remember it as a whole in memory (or if there is some contract around construction of those URLs maybe it would be good to extract it to separate responsibility class).

CHANGELOG.md
```diff
@@ -24,7 +24,13 @@ Adding a requirement of a major version of a dependency is breaking a contract.
 Dropping a requirement of a major version of a dependency is a new contract.
 
 ## [Unreleased]
-[Unreleased]: https://github.com/atlassian/jira-actions/compare/release-3.16.1...master
+[Unreleased]: https://github.com/atlassian/jira-actions/compare/release-3.17.0...master
+
+## [3.17.0] - 2022-04-13
```
pczuj, line 29: I don't think we released that. 

Our approach is to mark release in CANGELOG after it's released. Then it's certain what is inside it and when it was released.

You should add the `### Added` just to the `## [Unreleased]`

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/WebJira.kt
pczuj: I'd consider following similar naming of either "comment" or "issue comment" instead of "focused comment" in other places of the code

pczuj (approved):
I think that's it when it comes to my review of the current state of the PR. I'm approving as a sign of that, but please consider addressing my comments.

CHANGELOG.md
ewefie: I'll stick to the previous convention.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/ActionType.kt
ewefie: I'll stick to the existing convention.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/action/ViewHistoryTabAction.kt
ewefie: Good point, I'll add it.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/memories/Comment.kt
ewefie: I didn't think about comment memory in that way. Sure, users doesn't care about comment id. I've added it here because it was convenient to have this whole parsing html in one place (while creating Comment object). I'll consider moving comment id extraction to a different place, if it's not too complicated.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/WebJira.kt
ewefie: I agree. I was thinking about having View History Tab Action implemented as an extension to View Issue Action. Then, while being on issue view we could just click on the tab and measure the loading time but with this approach it would be hard to separate this actions from the existing ones. Maybe not in this PR, but later I'll come up with some improvement suggestions.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/WebJira.kt
ewefie: You are right about simplifying those names. I'll change it.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/JiraCoreScenario.kt
ewefie: Well, I'm also not sure if I should add new actions to existing scenario or create a new one. I don't think this may break something, but at the same time I'm not sure how (and if at all) this library is used besides JPT. Even if I go with creating a new scenario, there's still a change to the existing one - View Issue action got additional memory for comments. 
Maybe other reviewers have some recommendation? cc/ @lipinskipawel @dagguh @wyrzyk

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/CommentTabPanel.kt
```diff
@@ -0,0 +1,89 @@
...
+import org.openqa.selenium.support.ui.ExpectedConditions
+import java.time.Duration
+
+class CommentTabPanel(
+    val driver: WebDriver
+) {
+    val loadMoreCommentsJira8Locator: By = By.cssSelector("a.show-more-comments")
+    val loadMoreCommentsJira9Locator: By = By.cssSelector("button.show-more-comment-tabpanel")
+
+    fun waitForCommentTabpanelActive(): CommentTabPanel {
```
pczuj, line 17: Nit: 
```suggestion
    fun waitForCommentTabPanelActive(): CommentTabPanel {
```

pczuj: Or should it be `waitForActive()`? It's already inside `CommentTabPanel` class.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/CommentTabPanel.kt
```diff
@@ -0,0 +1,89 @@
...
+    fun getComments(): Set<Comment> {
+        val comments = driver.findElements(By.cssSelector("#issue_actions_container > div.activity-comment"))
+        return comments.map {
+            val id = it.getAttribute("id").split("-")[1]
+            val url = it.findElement(By.cssSelector(".comment-created-date-link")).getAttribute("href")
+            Comment(id, url)
+        }.toSet()
+    }
+
+    fun validateCommentIsFocused(
```
pczuj, line 78: WDYT about
```suggestion
    fun isCommentFocused(
```
?

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/action/ViewIssueAction.kt
```diff
@@ -16,7 +13,8 @@ class ViewIssueAction(
     private val meter: ActionMeter,
     private val issueKeyMemory: IssueKeyMemory,
     private val issueMemory: IssueMemory,
-    private val jqlMemory: JqlMemory
+    private val jqlMemory: JqlMemory,
+    private val commentMemory: CommentMemory
```
pczuj, line 17: Note: This is breaking change.

Do we want to maintain backwards compatibiity of `ViewIssueAction` constructors here?

pczuj: Solution suggestion: We could transition actions constructtion to use builders as we do in some other JPT modules.

pczuj: BTW there is a good tool for SemVer API comparison: https://siom79.github.io/japicmp/

It can be easily configured in a maven project like this:
```
<plugin>
    <groupId>com.github.siom79.japicmp</groupId>
    <artifactId>japicmp-maven-plugin</artifactId>
    <version>0.15.4</version>
    <configuration>
        <oldVersion>
            <dependency>
                <groupId>com.atlassian.performance.tools</groupId>
                <artifactId>aws-infrastructure</artifactId>
                <version>2.23.1</version>
            </dependency>
        </oldVersion>
        <newVersion>
            <dependency>
                <groupId>com.atlassian.performance.tools</groupId>
                <artifactId>aws-infrastructure</artifactId>
                <version>2.23.2-SNAPSHOT</version>
            </dependency>
        </newVersion>
        <parameter>
            <ignoreMissingClasses>true</ignoreMissingClasses>
            <includes>
                <include>com.atlassian.performance.tools.awsinfrastructure.api.*</include>
            </includes>
            <includeExclusively>true</includeExclusively>
        </parameter>
    </configuration>
    <executions>
        <execution>
            <phase>verify</phase>
            <goals>
                <goal>cmp</goal>
            </goals>
        </execution>
    </executions>
</plugin>
```
(needs to be adapted to specific project)

CHANGELOG.md
```diff
@@ -26,6 +26,10 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/jira-actions/compare/release-3.16.1...master
 
+### Added
```
pczuj, line 29: This PR is actually also:
 - Adding new methods to `WebJira`
 - Adding new page objects
 - Adding new memory
 - Appending `ViewIssueAction` with comment remembering capability

all of those are in the `api` package, so maybe they should all be mentioned in the changelog?

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/JiraCoreScenario.kt
dagguh: Extending the `JiraCoreScenario` coverage is valuable and compatible. Switching those tabs **is** a part of a Jira Core user scenario. I understand we might be afraid of introducing flakiness or bugs, but:
* we should quickly identify and fix those
* the consumer can roll back to old lib version

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/WebJira.kt
dagguh: Clicking on a tab can be inlined in the `ViewIssueAction` with a given probability.
Years ago, I used a side-repo for Cloud vs Server benchmark, which has such probabilistic sub-actions, like [WorkOnIssue](https://github.com/atlassian/jces-1209/blob/master/custom-vu/src/main/kotlin/jces1209/vu/action/WorkOnIssue.kt). Props to @atlassian/jpt-cloud team for extending the coverage so much :heart: 
It has an extra abstraction to hide the differences between Server and Cloud, but we could actually harvest that too (and potentially expand for DC or Jira 8/9).

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/action/ViewCommentAction.kt
```diff
@@ -0,0 +1,36 @@
...
+
+        meter.measure(
+            key = VIEW_COMMENT,
+            action = { jira.goToComment(comment.url).validateCommentIsFocused(comment.id) },
+            observation = { page ->
+                Json.createObjectBuilder()
+                    .add("issueKey", page.getIssueKey())
+                    .add("commentId", comment.id)
+                    .build()
+            }
```
dagguh, line 33: Do we have a use case for those observations?

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/action/ViewIssueAction.kt
dagguh: Yes, we have to maintain source and binary compatibility. Builder pattern is a great way to achieve that.

dagguh: WRT API checker, a PoC in JPT would be awesome.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/HistoryTabPanel.kt
```diff
@@ -0,0 +1,29 @@
...
+import java.time.Duration
+
+class HistoryTabPanel(
+    private val driver: WebDriver
+) {
+    private val loadMoreEntriesLocator = By.cssSelector("button.show-more-changehistory-tabpanel")
+
+    fun loadAllHistoryEntries(): HistoryTabPanel {
+        if (driver.isElementPresent(loadMoreEntriesLocator)) {
+            val loadMoreButton = driver.findElement(loadMoreEntriesLocator)
```
dagguh, line 15: BTW. You can `val loadMoreButton = driver.findElements(loadMoreEntriesLocator).singleOrNull()` and work with a nullable `loadMoreButton: WebElement?` rather than an `isElementPresent`. I believe that's true for vast majority of `isElementPresent` calls.

dagguh (commented):
Outstanding work, well done

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/JiraCoreScenario.kt
dagguh: What was the existing behavioral contract?
> Javadoc comment: `Provides Jira Core specific Scenario`
> class name `JiraCoreScenario`

Viewing focused comment and viewing history tab still fit. We never promised *not* to do more actions.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/action/ViewIssueAction.kt
pczuj: I created https://ecosystem.atlassian.net/browse/JPERF-804 for adding japicmp to our repositories, though I don't think I will have time to ship this soon.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/CommentTabPanel.kt
```diff
@@ -0,0 +1,89 @@
...
+        driver.findElement(loadMoreCommentsJira9Locator).click()
+
+        if (driver.isElementPresent(loadMoreCommentsJira8Locator) && driver.findElement(loadMoreCommentsJira9Locator)
+                .getAttribute("data-load-all-enabled").equals("true")
+        ) {
+            val loadMoreButton = driver.findElement(loadMoreCommentsJira9Locator)
+
+            Actions(driver)
+                .keyDown(Keys.SHIFT)
+                .click(loadMoreButton)
```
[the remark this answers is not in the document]
pczuj: Maybe weird, but seems to actually be the correct way
Screenshot 2022-05-11 at 15 35 16

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/action/ViewCommentAction.kt
pczuj: https://github.com/atlassian/jira-actions/pull/103#discussion_r864589103

Without issue key we won't be able to know if the view comment action performance depends on the issue viewed. Some issues may have more comments than others and some even can have content of those comments that could render longer. I think it's worth to add.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/memories/Comment.kt
```diff
@@ -0,0 +1,6 @@
+package com.atlassian.performance.tools.jiraactions.api.memories
+
+data class Comment(
```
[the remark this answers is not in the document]
ewefie: Makes sense. I just put it in the package where the rest of data classes for memories is stored. Wdyt about moving all of them to a new package?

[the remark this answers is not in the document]
dagguh: The package is right. This is an API, already required by `CommentMemory`, which is already required by `ViewCommentAction` (both also API).
The solution is to remove the `data` keyword and manually add `equals`/`hashCode` if necessary.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssuePage.kt
```diff
@@ -82,4 +82,12 @@ class IssuePage(
 
     fun getIssueId(): Long = driver.findElement(By.id("key-val")).getAttribute("rel").toLong()
     fun getIssueType(): String = driver.findElement(By.id("type-val")).text
+
+    internal fun openCommentTabPanel(): CommentTabPanel {
+        return CommentTabPanel(driver).waitForActive()
+    }
+
+    internal fun openHistoryTabPanel(): HistoryTabPanel {
```
pczuj, line 90: - [ ] Consider removing `internal`

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/memories/Comment.kt
pczuj: We discussed the matter via zoom call. We agreed that `id` can stay. This is mainly, so that we can proceed with the merge ASAP.

Jira's implementation details around this is highly unlikely to change and we are using it already when validating that the comment was focused.

CHANGELOG.md
pczuj: Because this is not necessary I'm closing this conversation.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/WebJira.kt
pczuj: This was addressed. Closing the converastion.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/action/ViewIssueAction.kt
pczuj: Builder was introduced and we reviewed it together with Ewelina yesterday on a zoom call.

We did also run japicmp on the changes and saw that there are only non-breaking changes introduced. I'm resolving the conversation.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/memories/Comment.kt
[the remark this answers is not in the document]
pczuj: `data` got removed and `equals`/`hashCode` was added. Resolving this conversation.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssuePage.kt
pczuj: This came up when we were reviewing japicmp report on a zoom call yesterday. The `internal` part got removed, so I'm closing the conversation and marking the task.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/action/ViewCommentAction.kt
pczuj: @dagguh is the use case acceptable?

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/CommentTabPanel.kt
```diff
@@ -0,0 +1,89 @@
...
+        )
+        return this
+    }
+
+    fun showAllComments(): CommentTabPanel {
+        if (driver.isElementPresent(loadMoreCommentsJira8Locator)) {
+            return showAllCommentsInJira8()
+        } else if (driver.isElementPresent(loadMoreCommentsJira9Locator)) {
+            return showAllCommentsInJira9()
+        }
```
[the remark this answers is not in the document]
pczuj: Probably `loadMoreCommentsJira8Locator` is a bit misleading as I believe it works similar in Jira 7. @ewefie could you confirm?

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/memories/CommentMemory.kt
```diff
@@ -0,0 +1,3 @@
+package com.atlassian.performance.tools.jiraactions.api.memories
+
+interface CommentMemory : Memory<Comment>
```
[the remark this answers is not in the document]
pczuj: @ewefie this is not addressed, what's your decision here?

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/CommentTabPanel.kt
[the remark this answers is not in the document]
dagguh: Now we've got an `else` :+1:, but `logger.debug("This Jira version is not supported.")` is a quiet fail. If Jira 10 comes out, breaks semantic HTML, then testers silently lose the `showAllComments` measurement without knowing it. We should throw.

PR #105, opened 2022-06-02
Migrate to artifact publish token

rinconjc:
Artifactory secrets are being deprecated.

Please review and merge this change to use the new artifact publish token.


pczuj:
Hi @rinconjc, thanks for your help. I will be looking at it and related PRs in other modules today.

Sorry for a bit late response.

pczuj (approved):
Ok, it took me some time to actually learn how those changes work, but I believe I understand those now.

PR #106, opened 2022-06-10
MNSTR-6059 Fix View History Tab Action by loading all entries in J9. …

ewefie:
In this PR I'm fixing issues with recently added actions: ViewHistoryTab and ViewComment.

View History Tab Action:
- in Jira 9 only the first batch of entries was loaded after clicking on 'show more' button, later the presence of the button was ignored. After fix clicking on 'show more' button is handled in the loop,
- in Jira 8 in some cases previously set duration was not enough to finish loading the entries. I increased it for all timed conditions to avoid timeout exceptions.

View Comment Action:
- improper selector was used in the function responsible for loading all comments in Jira 9,
- the previous logic for loading comments in Jira 9 was replaced with the same solution as for View History Tab - clicking 'show more' button in the loop,
- I increased the time duration in timed condition similarly as for View History Tab Action.

### DIFF
--- a/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/CommentTabPanel.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/CommentTabPanel.kt
@@ -15,16 +15,20 @@ class CommentTabPanel(
 ) {
     private val logger: Logger = LogManager.getLogger(this::class.java)
 
+    // the duration is set to such value because viewing issue with many comments is a heavy operation, and we assume
+    // the user is patient enough to wait for all comments to show up
+    private val duration: Duration = Duration.ofMinutes(2)
     private val loadMoreCommentsJira8Locator: By = By.cssSelector("a.show-more-comments")
     private val loadMoreCommentsJira9Locator: By = By.cssSelector("button.show-more-comment-tabpanel")
+    private val attemptLimit = 5
 
     fun waitForActive(): CommentTabPanel {
         driver.wait(
-            Duration.ofSeconds(45),
+            duration,
             ExpectedConditions.elementToBeClickable(By.id("comment-tabpanel"))
         ).click()
         driver.wait(
-            Duration.ofSeconds(60),
+            duration,
             ExpectedConditions.presenceOfElementLocated(By.cssSelector("#comment-tabpanel.active-tab"))
         )
         return this
@@ -43,7 +47,7 @@ class CommentTabPanel(
         commentId: String
     ): CommentTabPanel {
         driver.wait(
-            Duration.ofSeconds(30),
+            duration,
             ExpectedConditions.visibilityOfElementLocated(By.cssSelector("#comment-$commentId.focused"))
         )
         return this
@@ -57,7 +61,7 @@ class CommentTabPanel(
         } else if (driver.isElementPresent(loadMoreCommentsJira9Locator)) {
             return showAllCommentsInJira9()
         } else {
-            logger.debug("This Jira version is not supported.")
+            logger.debug("The number of comments is less than ten or this Jira version is not supported.")
         }
         return this
     }
@@ -66,20 +70,18 @@ class CommentTabPanel(
         val loadMoreButton = driver.findElement(loadMoreCommentsJira8Locator)
         loadMoreButton.click()
         driver.wait(
-            timeout = Duration.ofSeconds(60),
-            condition = ExpectedConditions.stalenessOf(loadMoreButton)
+            duration,
+            ExpectedConditions.stalenessOf(loadMoreButton)
         )
         return this
     }
 
     private fun showAllCommentsInJira9(): CommentTabPanel {
-        driver.findElement(loadMoreCommentsJira9Locator).click()
-
-        if (driver.isElementPresent(loadMoreCommentsJira8Locator) && driver.findElement(loadMoreCommentsJira9Locator)
-                .getAttribute("data-load-all-enabled").equals("true")
-        ) {
+        repeat(attemptLimit) {
+            if (!driver.isElementPresent(loadMoreCommentsJira9Locator)) {
+                return this
+            }
             val loadMoreButton = driver.findElement(loadMoreCommentsJira9Locator)
-
             Actions(driver)
                 .keyDown(Keys.SHIFT)
                 .click(loadMoreButton)
@@ -87,8 +89,8 @@ class CommentTabPanel(
                 .build()
                 .perform()
             driver.wait(
-                timeout = Duration.ofSeconds(60),
-                condition = ExpectedConditions.stalenessOf(loadMoreButton)
+                duration,
+                ExpectedConditions.stalenessOf(loadMoreButton)
             )
         }
         return this
--- a/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/HistoryTabPanel.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/HistoryTabPanel.kt
@@ -1,41 +1,61 @@
 package com.atlassian.performance.tools.jiraactions.api.page
 
 import org.openqa.selenium.By
+import org.openqa.selenium.Keys
 import org.openqa.selenium.WebDriver
+import org.openqa.selenium.interactions.Actions
 import org.openqa.selenium.support.ui.ExpectedConditions
 import java.time.Duration
 
 class HistoryTabPanel(
     private val driver: WebDriver
 ) {
+    // the duration is set to such value because viewing issue with many change history entries is a heavy operation,
+    // and we assume the user is patient enough to wait for all entries to show up
+    private val duration: Duration = Duration.ofMinutes(2)
     private val loadMoreEntriesLocator = By.cssSelector("button.show-more-changehistory-tabpanel")
+    private val attemptLimit = 5
 
-    fun loadAllHistoryEntries(): HistoryTabPanel {
+    fun showAllHistoryEntries(): HistoryTabPanel {
         if (driver.isElementPresent(loadMoreEntriesLocator)) {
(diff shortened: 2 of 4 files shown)


src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/CommentTabPanel.kt
```diff
@@ -20,11 +20,11 @@ class CommentTabPanel(
 
     fun waitForActive(): CommentTabPanel {
         driver.wait(
-            Duration.ofSeconds(45),
+            Duration.ofSeconds(120),
```
szymonra, line 23: I think Duration.ofMinutes(2) would be slightly more clear. Also if we are using this two minutes duration in most of the places I think it would be good to extract it to constant.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/CommentTabPanel.kt
```diff
@@ -87,10 +83,12 @@ class CommentTabPanel(
                 .build()
                 .perform()
             driver.wait(
-                timeout = Duration.ofSeconds(60),
```
szymonra, line 90: Is there a direct need to mark timeout and condition parameters? If not I think it will be better to stick with the convention used before - just passing the parameters. One way of another, it would be the best to stick with one convention, unless for some reason it is not sensible.

szymonra (dismissed):

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/CommentTabPanel.kt
```diff
@@ -17,14 +17,15 @@ class CommentTabPanel(
 
     private val loadMoreCommentsJira8Locator: By = By.cssSelector("a.show-more-comments")
     private val loadMoreCommentsJira9Locator: By = By.cssSelector("button.show-more-comment-tabpanel")
+    private val duration: Duration = Duration.ofMinutes(2)
```
dagguh, line 20: Would a use actually wait 2 minutes?

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/HistoryTabPanel.kt
```diff
@@ -1,41 +1,59 @@
...
+                .click(loadMoreButton)
+                .keyUp(Keys.SHIFT)
+                .build()
+                .perform()
+            driver.wait(
+                duration,
+                ExpectedConditions.stalenessOf(loadMoreButton)
+            )
+        } while (
+            driver.isElementPresent(loadMoreEntriesLocator)
```
pczuj, line 55: We should limit users patience here, so that we won't fall into infinite loop here (e.g. button not working due to network issue). I think that after 5-10 clicks I would give up.

pczuj: With adding of this patience I wonder if we shouldn't fail the action in case of not loading all of the comments.

Alternatively at least maybe we should add number of comments loaded to the `observation` parameter of `meter.measure()`. This would be to make it possible to ensure that we are comparing apples to apples, e.g. if one version if Jira, doesn't cheat by loading less data.

pczuj: Same for `CommentTabPanel#showAllCommentsInJira9`

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/CommentTabPanel.kt
szymonra: I still think it is a good practice and while I get your point, I see this case as slightly different from what you put in the example. We switched from a specific durations to the default one, which is the same for all of the usages. However you are right that the memory/performance impact may be insignificant and if you think that it will be better to leave it the way it was, to stress out the we should approach each of the timeouts present here separately, I'm fine with that.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/CommentTabPanel.kt
ewefie: If they want to see all comments then they have to wait several minutes, in some cases much more than 2, but I know  that usually users give up after few seconds. If I change the value to more realistic we would important information about how our product behave under high load.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/CommentTabPanel.kt
pczuj: Yes, it's good practice to deduplicate, however I claim that in this case there was no duplication, because timeouts were different for most of the operations. Even if accidentally some of them matched it wasn't because we treated it as the same value.

We could create a common default and name it e.g. `usersPatience`. This would allow us to document rationale behind the value, because as @dagguh pointed out there was research around it. However I don't think that in our case it has enough value to balance the potential risk. Deduplicating it would suggest to ourselves and future contributors that the timeouts are meant to be manipulated all at once and not with proper granularity.

dagguh: Yeah, gotta watch out for accidental duplication. Maybe some SSH timeout is also set to `Duration.ofMinutes(2)`, but it's semantically different, they shouldn't reuse the same symbol.
And don't worry about `Duration` constructor performance, it's instant (pun intended :wink:).

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/CommentTabPanel.kt
pczuj: I think that we can assume that the VU who is doing the performance measures for us is rather patient for the purpose of finishing the action even if it's slow. The VU should be still have limited patience, but it's purpose is a bit different then the regular user patience, which is to not get stuck.

I'm fine with high patience of the automation as long as we have processes to monitor it (and the test does that).

pczuj:
I'm approving with the adnotation that actions given by checkboxes in comments of this PR need to be addressed.

I believe we don't have automation to check that in GitHub.

pczuj (dismissed):

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/CommentTabPanel.kt
dagguh: Reasonable justification :+1: 
Let's share it via a code comment to make it obvious to everyone.

dagguh:
We can use the [`Require conversation resolution before merging` protection](https://github.com/atlassian/jira-actions/settings/branch_protection_rules/3773101)
image

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/CommentTabPanel.kt
ewefie: I have deduplicated duration values for both new actions, as every single step on those views seems to be very slow, but I didn't touch any other timed conditions.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/HistoryTabPanel.kt
ewefie: I like the idea of adding the number of loaded elements to the observation, but would it be okay for you to implement it later in a separate PR?

pczuj: @ewefie I'm specifically using "Consider..." in the action checkbox text whenever I'm ok with not doing this.

Feel free to skip this feedback if you think it's not valid for this PR.

If you think it could still be done later then you could create a [JPERF](https://ecosystem.atlassian.net/browse/JPERF) issue and address it later even by yourself.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/HistoryTabPanel.kt
```diff
@@ -1,41 +1,62 @@
...
+                .keyUp(Keys.SHIFT)
+                .build()
+                .perform()
+            driver.wait(
+                duration,
+                ExpectedConditions.stalenessOf(loadMoreButton)
+            )
+            attemptCount++
+        } while (
+            attemptCount < attemptLimit && driver.isElementPresent(loadMoreEntriesLocator)
```
pczuj, line 58: Nit: `for` loop with conditional `break` could be a bit more clean/safe here. Using `while` is always worrying me about deadlocks.

TIL: Apparently Kotlin doesn't have `break`, however `return` with labels can be used to achieve the same capability. This approach is compatible with other alternatives to simple `for` like `repeat()`: https://kotlinlang.org/docs/returns.html#return-to-labels

In this case it would be possible to simply `return this` in the scope of the full method - no labels needed.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/HistoryTabPanel.kt
```diff
@@ -1,41 +1,61 @@
...
         driver.wait(
-            Duration.ofSeconds(60),
+            duration,
             ExpectedConditions.presenceOfElementLocated(By.cssSelector("#changehistory-tabpanel.active-tab"))
         )
         return this
     }
+
+    private fun showMoreEntriesIfNeeded(): HistoryTabPanel {
+        repeat(5) {
```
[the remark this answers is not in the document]
pczuj: Maybe you even wanted to reuse the `attemptLimit` variable on the class level?

pczuj (dismissed):

pczuj (approved):

PR #108, opened 2022-07-20
JPERF-814: Correct issue comment link reading to work with Jira versions prior to 8.17.0

pczuj:


CHANGELOG.md
```diff
@@ -26,6 +26,11 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/jira-actions/compare/release-3.17.2...master
 
+### Fixed
+- Correct issue comment link reading to work with Jira versions prior to 8.17.0. Fix [JPERF-814].
```
dagguh, line 30: Hmm, but [CI is green for 8.5.0 on master](https://github.com/atlassian/jira-actions/runs/7087593863?check_suite_focus=true).

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/CommentTabPanel.kt
```diff
@@ -38,7 +38,7 @@ class CommentTabPanel(
         val comments = driver.findElements(By.cssSelector("#issue_actions_container > div.activity-comment"))
         return comments.map {
             val id = it.getAttribute("id").split("-")[1]
-            val url = it.findElement(By.cssSelector(".comment-created-date-link")).getAttribute("href")
+            val url = it.findElement(By.cssSelector(""".action-head a[href^="/browse/"]""")).getAttribute("href")
```
dagguh, line 41: We can add Jira version support as a behavioral contract (of `getComments` or `CommentTabPanel` or the whole lib).

CHANGELOG.md
pczuj: Yup, this indeed is strange. Needs investigation.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/CommentTabPanel.kt
pczuj: What do you mean?

dagguh:
```java
/**
 * Represents an issue comment tab panel from Jira 8.5.0-8.17.0.
 */
class CommentTabPanel(
```

CHANGELOG.md
pczuj: I'm gonna debug `JiraCoreScenarioIT` and see why the problem doesn't occur.

For more context: The problem started to happen inside `JiraSoftwareScenarioIT` when I bumped jira-actions in jira-software-actions and run it's tests locally.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/CommentTabPanel.kt
dagguh: Fair enough

CHANGELOG.md
pczuj: Bad new is that the `JiraCoreScenarioIT` doesn't seem to test anything: https://ecosystem.atlassian.net/browse/JPERF-815

Good news is that we know about it now and we can fix it.

pczuj: I will push a branch proving the issue

pczuj: Pushed here https://github.com/atlassian/jira-actions/pull/109

dagguh: Whoa! :exploding_head: Great catch!

pczuj: Interesting idea. 

I was also considering contributing a `JqlRememberingProjectMemory` which would be built from both `ProjectMemory` and `JqlMemory`, so that when remembering a project it also would remember JQL like this
```
project = ${project.key} order by issuekey desc
``` 

I'm personally using such JQLs frequently. Possibly typical Jira user too?

A code for this memory is very simple
```
package com.atlassian.performance.tools.jiraactions.api.memories.adaptive

import com.atlassian.performance.tools.jiraactions.api.memories.JqlMemory
import com.atlassian.performance.tools.jiraactions.api.memories.Project
import com.atlassian.performance.tools.jiraactions.api.memories.ProjectMemory

class JqlRememberingProjectMemory(
    private val delegate: ProjectMemory,
    private val jqlMemory: JqlMemory
) : ProjectMemory {
    override fun recall() = delegate.recall()

    override fun remember(memories: Collection<Project>) = delegate.remember(memories)
        .also {
            jqlMemory.remember(memories.map { "project = ${it.key} order by issuekey desc" })
        }
}
```

dagguh: Ship it :rocket:

pczuj:
Pushed changes without fix. CI should fail with expected IssueView incompatibility.

When tests will finish I will push a commit with fix, which is already prepared (no changes from the original PR)

pczuj:
> org.openqa.selenium.NoSuchElementException: no such element: Unable to locate element: {"method":"css selector","selector":".comment-created-date-link"}

As expected. Though I wanted all jobs to finish with this, yet after one failed the rest were cancelled :(

pczuj:
Pushed the actual fix 🤞

pczuj:
> Though I wanted all jobs to finish with this, yet after one failed the rest were cancelled :(

Just pushed a fix for that.

I think it makes no sense to fail fast. We will gather more information about the failure and on what versions is happens.

.github/workflows/ci.yml
```diff
@@ -15,6 +15,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
```
dagguh, line 18: Nice, today I learned

dagguh (approved):

pczuj:
New test failures are weird and interesting. I can reproduce it easily on local.

Based on the recording I believe it might be something about not waiting for page load to happen after disabling RTE. Could it be that 2x `driver.navigate().to(base.resolve(path).toURL())` executed without wait will make the 2nd one to not happen?

pczuj:
I'm failing to understand why the failure of the `SearchJqlAction` happened. I raised an issue for it: https://ecosystem.atlassian.net/browse/JPERF-818 and pasted to it a recording of the problem.

It's weird that I'm able to reproduce the failure without any changes on top of current state of this PR, but when I'm experimenting around it it's not happening.

I tried to loop over 100x disable RTE -> SearchJqlAction (in fact: enable RTE -> disable RTE -> SearchJqlAction). Everything works fine.

I also tried to do 2x `driver.navigate().to(base.resolve(path).toURL())` without wait. Even in the loop of 100x it just ends on the 2nd path, so not as I thought.

I don't really have a good idea what else can I do to diagnose the problem.

pczuj:
Actually I've got one idea. Maybe it's because of the redirect happening after disabling RTE.

I will check if doing 100x: 
1. Navigate to /secure/MyJiraHome.jspa
2. Navigate to /issues/?jql=
will always end up on /issues/?jql=

dagguh:
Which fails are you describing? When looking at 8.0.0, it fails due to `docker-infrastructure`:
```
Caused by: org.openqa.selenium.TimeoutException: Expected condition failed: waiting for visibility of element located by By.className: subnavigator-title (tried for 600 second(s) with 50 milliseconds interval)
	at org.openqa.selenium.support.ui.WebDriverWait.timeoutException(WebDriverWait.java:95)
	at org.openqa.selenium.support.ui.FluentWait.until(FluentWait.java:272)
	at com.atlassian.performance.tools.dockerinfrastructure.jira.SetUpFromScratchActionKt.wait(SetUpFromScratchAction.kt:179)
	at com.atlassian.performance.tools.dockerinfrastructure.jira.SetUpFromScratchActionKt.wait$default(SetUpFromScratchAction.kt:173)
	at com.atlassian.performance.tools.dockerinfrastructure.jira.SetUpFromScratchAction.cleanErrorMessages(SetUpFromScratchAction.kt:128)
	at com.atlassian.performance.tools.dockerinfrastructure.jira.SetUpFromScratchAction.createSampleProject(SetUpFromScratchAction.kt:124)
	at com.atlassian.performance.tools.dockerinfrastructure.jira.SetUpFromScratchAction.run$docker_infrastructure(SetUpFromScratchAction.kt:29)
	at com.atlassian.performance.tools.dockerinfrastructure.api.jira.AbstractJiraFormula.provisionJira(AbstractJiraFormula.kt:85)
	... 49 more
Caused by: org.openqa.selenium.NoSuchElementException: no such element: Unable to locate element: {"method":"css selector","selector":".subnavigator\-title"}
```

pczuj:
> Which fails are you describing?

https://github.com/atlassian/jira-actions/runs/7446215898?check_suite_focus=true
https://github.com/atlassian/jira-actions/runs/7446215981?check_suite_focus=true

pczuj:
Pushed 2 commits. 

First one is with a fix to a https://ecosystem.atlassian.net/browse/JPERF-819. It's a next problem which was hiding under fixes of `JiraCoreScenarioIT` (JPERF-815 and JPERF-816). The fix was discussed with @ewefie and seems to be fine.

The 2nd commit is about enabling 20 repeats of the `JiraCoreScenarioIT`. I did that to show that the https://ecosystem.atlassian.net/browse/JPERF-818 is pretty consistant.

Next after the failure will happen with the 20 retries i will push a fix to JPERF-818 that I can't prove to be fixing the flakiness other than showing that the 20 retries pass.

After that I will remove the 20 retries. The test executes for hours with it and we don't really want that.
Maybe we could have a switch in gradle that enables "hard mode", but I don't see this complexity necessary for now.

pczuj:
Hmmm... didn't expect the 7.2.0 CI to [succeed](https://github.com/atlassian/jira-actions/runs/7496808991?check_suite_focus=true). That's interesting... Possibly the flainess around JQL search action was introduced with certain Jira version and wasn't there when action was created originally. Nevertheless, my local tests on 8.5.3 suggested that a 2 line fix helps, so I'm gonna push it now to see if none of the total 100 runs will fail.

pczuj:
With the fix 100% of the CI is green with "hard mode" ([`JiraCoreScenarioIT` repeated 20 times](https://github.com/atlassian/jira-actions/pull/108/commits/e826f010c3f71ecc017a92a0a9a775e26f0d86e1))

dagguh (approved):
Well done! Nice and methodical

pczuj:
Removed the commit with "hard mode" on `JiraCoreScenarioIT` (20x repeats).

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/JiraCoreScenario.kt
```diff
@@ -20,8 +20,13 @@ class JiraCoreScenario constructor() : Scenario {
 
     override fun getActions(jira: WebJira, seededRandom: SeededRandom, meter: ActionMeter): List<Action> {
         initializeIssueKeyMemory(seededRandom)
-        val projectMemory = AdaptiveProjectMemory(random = seededRandom)
         val jqlMemory = AdaptiveJqlMemory(seededRandom)
+        val projectMemory = JqlRememberingProjectMemory
```
pczuj, line 24: One thing that I'm worried about is that with this the JQL memory will be flooded with "project = " queries. I guess the complicated logic inside the `AdaptiveJqlMemory` is what is supposed to prevent that, however the `JqlRememberingProjectMemory` ommits it.

pczuj: I added the `LimitedJqlMemory` to address that. See CHANGELOG.md and it's code to understand what it actually is and please make any suggetions if it could be better in any way.

pczuj:
I didn't add unit tests for `JqlRememberingProjectMemory` and for `LimitedJqlMemory` yet. I'm planning to do that, however thought that it might be worth to already push the `LimitedJqlMemory` implementation, as I'm currently not intending to change it.

Any suggestions about those 2 classes are welcome.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/memories/adaptive/LimitedJqlMemory.kt
```diff
@@ -0,0 +1,18 @@
+package com.atlassian.performance.tools.jiraactions.api.memories.adaptive
+
+import com.atlassian.performance.tools.jiraactions.api.memories.JqlMemory
+
+class LimitedJqlMemory(
```
mgrzaslewicz, line 5: It would be great to have a unit test for this class

mgrzaslewicz (dismissed):

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/CommentTabPanel.kt
pczuj, line 41: I'm trying to finish this PR after a while.

Taking a look again at this change I'm not sure if this selector change is moving the code in the right direction. The previous selector didn't work with older Jira versions, however was probably more "semantic". New selector has an assumption about contract which will unlikely change, but it's still some knowledge which probably shouldn't be here.

I'm thining if it wouldn't be better to somehow query this element both ways and select the first match. Not sure how to do that without compromising the readability and noise.

CC: @dagguh

mgrzaslewicz: @pczuj Maybe something like this? 
```
val jiraVersionsPriorTo817Selector = ".comment-created-date-link"
val jiraVersionsSelectorAfter817 = """.action-head a[href^="/browse/"]"""
val url = it.findElement(By.cssSelector("$jiraVersionsPriorTo817Selector, $jiraVersionsSelectorAfter817")).getAttribute("href")
```

.github/workflows/ci.yml
mgrzaslewicz, line 18: why not fail fast? A comment in the code would be great here

CHANGELOG.md
```diff
@@ -26,6 +26,22 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/jira-actions/compare/release-3.17.2...master
 
+### Added
+- Introduce `JqlRememberingProjectMemory` so that we have 2nd source of new JQL queries other than using `ViewIssueAction`, which already relies on `JqlMemory`.
+- Add `ProjectJqlFactory` and `SimpleProjectJqlFactory` primarily as components of the new `JqlRememberingProjectMemory`.
+- Introduce `LimitedJqlMemory` to allow for limiting number of queries remembered from specific source.
```
mgrzaslewicz, line 32: Could you add a reason for adding LimitedJqlMemory here or to its javadoc? Why do we need to limit number of queries?

.github/workflows/ci.yml
```diff
@@ -15,6 +15,7 @@ jobs:
...
       matrix:
         jira-version: ["7.2.0", "7.13.0", "8.0.0", "8.5.0", "8.13.0"]
```
mgrzaslewicz, line 20: Do we need all the old versions? Maybe we could start with 8.13 and add 8.20, 9.3

.github/workflows/ci.yml
mgrzaslewicz: Why do we need it? Please add a comment to the yml file

pczuj: I will add more rationale to the commit message, would that be good enough?

As a general rule I'm trying to avoid non-functional parts of code as those tend to get outdated. An exception of this rule is the documentation of the externally consumable contract (e.g. API docs).

CHANGELOG.md
pczuj: The rationale is in [the commit message](https://github.com/atlassian/jira-actions/pull/108/commits/a398a9e1ce3ca6701516199d1cbf98994400bc53). I could add it to the CHANGELOG, however I was under impression that I'm already adding to much of context in there.

.github/workflows/ci.yml
pczuj: Maybe we don't need them, however other changes in this PR don't need those to be changed, so I will leave them as they are.

pczuj: I actually could add 8.20.0 and 9.0.0 to reveal compatibility

pczuj:
Added commit that extends matrix of tested Jira versions and changed commit message for disabling `fail-fast` to include rationale.

mgrzaslewicz (approved):

pczuj:
Updated docker-infrastructure to pull in https://github.com/atlassian/docker-infrastructure/pull/18. Without it tests on Jira 8.20.0 and 9.0.0 are failing.

mgrzaslewicz (approved):

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/CommentTabPanel.kt
dagguh: Another way to "or":
```kotlin
val url = it.findElements(By.cssSelector(".comment-created-date-link"))
    .plus(it.findElements(By.cssSelector(".action-head a[href^='/browse/']")))
    .first()
    .getAttribute("href")
```
But I like @mgrzaslewicz's version, because it explains the 8.1.7 reasoning.

dagguh: There's some semantic misalignment.

The `.action-head` - "action" is old Jira wording for comment, so that's okayish, but "head" means the comment header. Header is a location, not a meaning. We're interested in the comment url/link, wherever it is located.

The `href^="/browse/"` assumes the structure of `Comment.url`, but this prefix is not comment-specific. If we need to assume the `href`, we should go for `*#comment*` or `*com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel*`.

The old `.comment-created-date-link` was quite good semantically, it's about a "comment link". It's unnecessarily polluted by "created date", but we can live with that.

Could you post a HTML snippet which doesn't work with `.comment-created-date-link`? It could be the case that the `.action-head` + href is the lesser of all evils, but let's see.

pczuj: In Jira 8.13.0 rendered comment HTML looks like this:
```
<div id="comment-10001" class="issue-data-block activity-comment twixi-block  expanded">
    <div class="twixi-wrap verbose actionContainer">
        <div class="action-head">
            <a href="#" title="Collapse comment" class="twixi"><span class="icon-default aui-icon aui-icon-small aui-iconfont-expanded"><span>Hide</span></span></a>
            <div class="action-links">
                <a href="/browse/FIR-10?focusedCommentId=10001&amp;page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-10001" title="Right click and copy link for a permanent link to this comment." class="activitymodule-link issue-comment-action">
                    <span class="icon-default aui-icon aui-icon-small aui-iconfont-link">Permalink</span></a>
                <a id="edit_comment_10001" href="/secure/EditComment!default.jspa?id=10009&amp;commentId=10001" title="Edit" class="edit-comment issue-comment-action">
                    <span class="icon-default aui-icon aui-icon-small aui-iconfont-edit">Edit</span></a>
                <a id="delete_comment_10001" href="/secure/DeleteComment!default.jspa?id=10009&amp;commentId=10001" title="Delete" class="delete-comment issue-comment-action">
                    <span class="icon-default aui-icon aui-icon-small aui-iconfont-delete">Delete</span></a>
            </div>
            <div class="action-details">





                <a class="user-hover user-avatar" rel="admin" id="commentauthor_10001_verbose" href="/secure/ViewProfile.jspa?name=admin"><span class="aui-avatar aui-avatar-xsmall"><span class="aui-avatar-inner"><img src="https://www.gravatar.com/avatar/e64c7d89f26bd1972efa854d13d7dd61?d=mm&amp;s=16" alt="admin"></span></span> admin</a>
                added a comment  - <span class="commentdate_10001_verbose subText"><span class="date user-tz" title="07/Oct/22 7:37 AM"><time class="livestamp" datetime="2022-10-07T07:37:41+0200">1 week ago</time></span></span>  </div>
        </div>
        <div class="action-body flooded"><p>Joined Sample Sprint 2 7 days 9 hours 10 minutes ago</p> </div>
    </div>
    <div class="twixi-wrap concise actionContainer">
        <div class="action-head">
            <a href="#" title="Expand comment" class="twixi"><span class="icon-default aui-icon aui-icon-small aui-iconfont-collapsed"><span>Show</span></span></a>
            <div class="action-details flooded">






                <a class="user-hover user-avatar" rel="admin" id="commentauthor_10001_concise" href="/secure/ViewProfile.jspa?name=admin"><span class="aui-avatar aui-avatar-xsmall"><span class="aui-avatar-inner"><img src="https://www.gravatar.com/avatar/e64c7d89f26bd1972efa854d13d7dd61?d=mm&amp;s=16" alt="admin"></span></span> admin</a>
                added a comment  - <span class="commentdate_10001_concise subText"><span class="date user-tz" title="07/Oct/22 7:37 AM"><time class="livestamp" datetime="2022-10-07T07:37:41+0200">1 week ago</time></span></span>                    Joined Sample Sprint 2 7 days 9 hours 10 minutes ago              </div>
        </div>
    </div>
</div>
```

pczuj:
After locking new dependencies the test on Jira 7.13.0 fails when viewing some content of SAM-5 issue. 
Screenshot 2022-10-17 at 17 32 40

I didn't check that, but I believe the of number of attachements (5) which makes the issue page higher for this issue combined with the license warning at the top of the page may be the reason why it doesn't work.

A proper fix probably would be to make the action work with the license warning, however I'm worried to invest time into this if the problem affects only legacy Jira 7.13.0.

Any suggestions what should we do?

pczuj:
I did rerun the whole CI 2 additional times:
 * The first one passed as a whole
 * The second one revealed that the symptom encountered on 7.13.0 can also happen with 7.2.0: https://github.com/atlassian/jira-actions/actions/runs/3250289873/jobs/5385009956

I wonder if it can also happen with any Jira 8/9 version.

pczuj:
The tests run on CI are still flaky yet we decided that we will address it after merging this PR to not block other contributions. In the end the flaky test wasn't running anyway, so the change improves the overall situation.

PR #112, opened 2022-12-13
JPERF-903 Open NotificationPopUps class to make it accessible from other JPT related libraries

ewefie:
To fix the flakiness of `edit sprint` action defined in `jira-software-actions` module I needed a way to close AUI flags. The mechanism for that was already implemented in this module however, `NotificationPopUps` was defined as an internal class. To make it possible to use this class in `jira-software-action` and avoid code duplication I removed that modifier.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,9 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/jira-actions/compare/release-3.17.3...master
 
+### Changed
+- Remove `internal` modifier from `NotificationPopUps`. Unblock [JPERF-903]
+
 ## [3.17.3] - 2022-10-19
 [3.17.3]: https://github.com/atlassian/jira-actions/compare/release-3.17.2...release-3.17.3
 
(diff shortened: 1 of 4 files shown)


mgrzaslewicz (dismissed):

src/main/kotlin/com/atlassian/performance/tools/jiraactions/page/NotificationPopUps.kt
```diff
@@ -9,7 +9,7 @@ import org.openqa.selenium.support.ui.ExpectedConditions.invisibilityOfElementLo
 import java.time.Duration
 
 
-internal class NotificationPopUps(private val driver: WebDriver) {
```
dagguh, line 12: According to the `CHANGELOG.md`, it also needs to be in the `api` subpackage

dagguh (changes requested):

mgrzaslewicz (approved):

dagguh (approved):

dagguh:
CI flaked for 7.13.0.
PS. Imagine having monorepo, and having to test `jira-software-actions`, `aws-infrastructure` and all other CIs. The combined flakiness, the duration, the combinations of test parameters like Jira versions.

PR #117, opened 2023-03-23
JPERF-1043 Add support for list view on IssueNavigatorPage. Add Issue…

ewefie:
Add support for list view on IssueNavigatorPage:
- Add IssueNavigatorView enum class that represents both available layouts and use it as an optional parameter in SearchJqlSimpleAction and SearchJqlSimpleAction constructors. When no value is provided, the detail view will be used as a default layout.
- While visiting the IssueNavigatorPage current view is detected and changed to the desired view if needed.

### DIFF
--- a/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/action/SearchIssues.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/action/SearchIssues.kt
@@ -15,6 +15,7 @@ import com.atlassian.performance.tools.jiraactions.api.page.issuenav.DetailView
 import com.atlassian.performance.tools.jiraactions.api.page.issuenav.IssueNavResultsView
 import org.apache.logging.log4j.LogManager
 import org.apache.logging.log4j.Logger
+import javax.json.Json
 import javax.json.JsonObject
 
 class SearchIssues private constructor(
@@ -43,8 +44,30 @@ class SearchIssues private constructor(
 
     private fun search(jqlQuery: String) = jira
         .goToIssueNavigator(jqlQuery)
-        .also { desiredView.switchToView() }
+        .also { switchResultsView() }
         .waitForResults(desiredView)
+
+    private fun switchResultsView() {
+        meter.measure(
+            switchActionType,
+            {
+                val observation = SwitchObservation.Builder(desiredView.javaClass.simpleName)
+                val selectedBefore = desiredView.isSelected()
+                observation.selectedBefore(selectedBefore)
+                if (!selectedBefore) {
+                    desiredView.switchToView()
+                    observation.switched()
+                }
+                observation
+            },
+            { observation ->
+                observation
+                    .selectedAfter(desiredView.isSelected())
+                    .build()
+                    .serialize()
+            }
+        )
+    }
 
     private fun observe(page: IssueNavigatorPage): JsonObject {
         val issueKeys = desiredView.listIssueKeys()
@@ -70,4 +93,42 @@ class SearchIssues private constructor(
 
         fun build(): Action = SearchIssues(jira, meter, actionType, jqlMemory, issueKeyMemory, desiredView)
     }
+
+    private val switchActionType = ActionType("Switch issue nav view") { SwitchObservation(it) }
+
+    private class SwitchObservation(
+        val desiredView: String,
+        val selectedBefore: Boolean,
+        val selectedAfter: Boolean,
+        val switched: Boolean
+    ) {
+        constructor(json: JsonObject) : this(
+            json.getString("desiredView"),
+            json.getBoolean("selectedBefore"),
+            json.getBoolean("selectedAfter"),
+            json.getBoolean("switched")
+        )
+
+        fun serialize(): JsonObject = Json.createObjectBuilder()
+            .add("desiredView", desiredView)
+            .add("selectedBefore", selectedBefore)
+            .add("selectedAfter", selectedAfter)
+            .add("switched", switched)
+            .build()
+
+        class Builder(
+            private var desiredView: String
+        ) {
+            private var selectedBefore: Boolean = false
+            private var selectedAfter: Boolean = false
+            private var switched: Boolean = false
+
+            fun selectedBefore(selectedBefore: Boolean) = apply { this.selectedBefore = selectedBefore }
+            fun selectedAfter(selectedAfter: Boolean) = apply { this.selectedAfter = selectedAfter }
+            fun switched() = apply { this.switched = true }
+
+            fun build() = SwitchObservation(desiredView, switched, selectedBefore, selectedAfter)
+        }
+    }
+
 }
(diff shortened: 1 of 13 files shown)


src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssueNavigatorPage.kt
```diff
@@ -6,40 +6,63 @@ import com.atlassian.performance.seleniumjs.NativeExpectedConditions.Companion.p
...
+        val currentLayout = getCurrentLayout()
+        if (currentLayout != view) {
+            driver.wait(elementToBeClickable(By.id("layout-switcher-button"))).click()
+            driver.wait(elementToBeClickable(By.cssSelector("a[data-layout-key=${view.selector}]"))).click()
+        }
+        return this
+    }
+
+    private fun getCurrentLayout(): IssueNavigatorView {
+        driver.wait(Duration.ofSeconds(10), presenceOfElementLocated(By.className("results-panel")))
```
ewefie, line 55: Unfortunately, to detect the currently selected view we need to wait for search results to appear.

dagguh: Hmm, we're using the sequence of `switchLayoutTo` and then `waitForIssueNavigator`. So if `By.className("results-panel")` is earlier than the `waitForIssueNavigator` conditions, it will not affect the measurements. In other words, if `waitForIssueNavigator` was already waiting for `results-panel` (indirectly), then wer're good.

If not, then we could do the switch *after* the page load. It would distort the proportions of the layouts, but it would preserve the measurement, which is more important.

dagguh: So I checked on dogfooding and indeed `results-panel` is a container for all the selectors in `val detailView` and `val listView`. So we're good.

But, the sequence is  `goToIssueNavigator(jqlQuery).switchLayoutTo(view).waitForIssueNavigator()`. So the `goToIssueNavigator` already performs the search, then `switchLayoutTo` can interrupt it, switch layout and re-trigger the search. It means that sometimes we'll make 1,5 search, affecting measurements.

So we can:
* bite the bullet and count on a low switch count
* switch after search
* prevent early search (no `jql=` in URI)

The `goToIssueNavigator` triggers JQL, because historically there was no other way. Going to "search for issues" was sending an empty JQL, which was interpreted as "all issues".
image
Since then, Jira changed empty JQL to mean "no search". So technically we can now `goToIssueNavigator` without a JQL, switch layout and then search.

dagguh (changes requested):
Good stuff overall, but let's discuss the measurement implications

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssueNavigatorPage.kt
ewefie: I chose this order of operations because, as you have noticed, to detect the currently selected layout, we need the `results-panel` to be present. If we reverse the order, there are some questions that arise: 
Should changing the layout be included in the measurement? If so, shouldn't we repeat the verification which is currently being done in the `waitForIssueNavigator` function at the end? If not, do we even need to change the layout at all? 
I must admit that I didn't consider the option of separating the search from actually going to the Issue Search view, perhaps that option is the best...

pczuj:
I don't have anything more to add to the review. Waiting with approve for application of suggestions.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssueNavigatorPage.kt
```diff
@@ -6,40 +6,63 @@ import com.atlassian.performance.seleniumjs.NativeExpectedConditions.Companion.p
...
+                emptyResults,
                 jiraErrors.anyCommonErrorNative()
             )
         )
         return this
     }
 
+    fun switchLayoutTo(view: IssueNavigatorView): IssueNavigatorPage {
+        val currentLayout = getCurrentLayout()
+        if (currentLayout != view) {
```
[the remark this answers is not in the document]
ewefie: Sure, it was a bit hard to pick one as both names are used in Jira, i.e.: we have "split-view" (sometimes called "detail-view") and "list-view" but to switch between them you have to click on `layout-switcher-button` an choose the option with proper `data-layout-key` 🙄 .

dagguh:
Feedback addressed in #118

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssueNavigatorPage.kt
dagguh: Switching views should be measured separately from searching issues. We can optimistically assume that switches won't be that frequent, if the scenario doesn't mix both views in one action list. But even then we can expect a lot of switches when the same Jira user is reused across VUs.

So the optimistic assumption must come with a verification - counting the switches. I'll just make a nested action for the switching part. It will have to inform us about each switch attempt and whether it actually switched or just detected that the view is already there. I'll do that in my PR->PR (#118).

dagguh: Done in https://github.com/atlassian/jira-actions/pull/118/commits/0acb5fc123efd40c4e1b270e40e56958c8c1f9dc

dagguh:
Red CI, investigating...

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/action/SearchIssues.kt
```diff
@@ -0,0 +1,133 @@
...
+        fun issueKeyMemory(issueKeyMemory: IssueKeyMemory) = apply { this.issueKeyMemory = issueKeyMemory }
+        fun desiredView(desiredView: IssueNavResultsView) = apply { this.desiredView = desiredView }
+
+        fun build(): Action = SearchIssues(jira, meter, actionType, jqlMemory, issueKeyMemory, desiredView)
+    }
+}
+
+private val switchActionType = ActionType("Switch issue nav view") { SwitchObservation(it) }
+
+private class SwitchObservation(
```
pczuj, line 100: If this is a private entity under scope of `SearchIssues` (it's in the `SearchIssues.kt` file), shouldn't it be inside the `SearchIssues` class?

dagguh:
Pushed in PR fixes and a rebase onto #120. After #120 merge, the commit list should be shorter.

dagguh:
All PR feedback is addressed (resolved conversations). We just wait for #120, rebase, green CI.

pczuj (approved):

dagguh:
#120 merged, this rebased onto master, awaiting CI...

pczuj:
#120 is merged

dagguh:
CI doesn't propagate to checks, another [incident](https://www.githubstatus.com/incidents/z3c6q056q332) :grimacing:

PR #118, opened 2023-03-24
JPERF-1043: list view support - addressing code review

dagguh:
My feedback from #117

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,12 +27,17 @@ Dropping a requirement of a major version of a dependency is a new contract.
 [Unreleased]: https://github.com/atlassian/jira-actions/compare/release-3.18.1...master
 
 ### Added
-- Support `ListView`, `DetailView` and other possible `IssueNavResultsView`s. Fix [JPERF-1043].
-- Add `IssueNavResultsView` parameter to JQL actions constructors.
+- Support `ListView`, `DetailView` and other possible `IssueNavResultsView`s.
+- Add `SearchIssues` action, which supports various `IssueNavResultsView`s. Fix [JPERF-1043].
 
 ### Deprecated
 - Deprecate most methods of `IssueNavigatorPage`,
   because they implicitly assumed `DetailView`, which won't work on `ListView`.
+- Deprecate all copies of `SearchJqlAction` (in favour of `SearchIssues`):
+  - `SearchJqlAction` itself
+  - `SearchJqlChangelogAction`
+  - `SearchJqlSimpleAction`
+  - `SearchJqlWildcardAction`
 
 ### Fixed
 - Randomly pick between `ListView` and `DetailView` in `JiraCoreScenario`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,7 @@ Dropping a requirement of a major version of a dependency is a new contract.
 
 ### Added
 - Support both list view and detail view in `IssueNavigatorPage`. Fix [JPERF-1043].
-- Add `IssueNavigatorView` parameter to JQL actions constructors.
+- Add `IssueNavigatorPage.View` parameter to JQL actions constructors.
 
 ### Fixed
 - Randomly pick between list view and detail view in `JiraCoreScenario`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,10 +27,13 @@ Dropping a requirement of a major version of a dependency is a new contract.
 [Unreleased]: https://github.com/atlassian/jira-actions/compare/release-3.18.1...master
 
 ### Added
-- Add optional `IssueNavigatorView` parameter to JQL actions constructors. Make `IssueNavigationPage` support both list view and split view. Fix [JPERF-1043]. 
+- Support both list view and detail view in `IssueNavigatorPage`. Fix [JPERF-1043].
+- Add `IssueNavigatorView` parameter to JQL actions constructors.
 
 ### Fixed
 - Randomly pick between list view and detail view in `JiraCoreScenario`.
+
+[JPERF-1043]: https://ecosystem.atlassian.net/browse/JPERF-1043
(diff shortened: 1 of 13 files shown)


CHANGELOG.md
```diff
@@ -27,7 +27,20 @@ Dropping a requirement of a major version of a dependency is a new contract.
...
+- Add `IssueNavResultsView` parameter to JQL actions constructors.
+
+### Deprecated
+- Deprecate most methods of `IssueNavigatorPage`,
+  because they implicitly assumed `DetailView`, which won't work on `ListView`.
+- Deprecate all copies of `SearchJqlAction` (in favour of `SearchIssues`):
+  - `SearchJqlAction` itself
+  - `SearchJqlChangelogAction`
+  - `SearchJqlSimpleAction`
+  - `SearchJqlWildcardAction`
```
dagguh, line 41: Yes, `SearchIssues` is more code now, but it allows us to delete all that code in the next major release.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/issuenav/IssueNavResultsView.kt
```diff
@@ -0,0 +1,26 @@
+package com.atlassian.performance.tools.jiraactions.api.page.issuenav
+
+import com.atlassian.performance.seleniumjs.NativeExpectedCondition
+
+interface IssueNavResultsView {
```
pczuj, line 5: Do we want this interface to be our SPI (in API)? An alternative could be an selection type similar to `enum`, so that we could hide the methods of this interface from API. 

My assumption is that we don't want the consumer to implement their own versions of this interface. Probably Jira has limited number of different search views, so I guess the lib has the implementation list complete. If that's the case we don't need to expose the messaging between implementations of `IssueNavResultsView` and `SearchIssues`.

On the other hand I'm actually a fan of designing software that exposes everything that it does instead of one that has a lot of room to do hidden stuff to achieve outcome. I personally believe that the consumer should have all the power to define composition layer of the application and rebuild the dependencies to work (or not work) as they want. Default composition should be enough to limit configuration complexity.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/issuenav/DetailView.kt
```diff
@@ -0,0 +1,53 @@
...
+import com.atlassian.performance.seleniumjs.NativeExpectedConditions.Companion.presenceOfElementLocated
+import com.atlassian.performance.tools.jiraactions.api.page.isElementPresent
+import com.atlassian.performance.tools.jiraactions.api.page.wait
+import com.atlassian.performance.tools.jiraactions.api.webdriver.JavaScriptUtils
+import org.openqa.selenium.By
+import org.openqa.selenium.WebDriver
+import org.openqa.selenium.support.ui.ExpectedConditions.elementToBeClickable
+import java.time.Duration
+
+class DetailView(
```
pczuj, line 15: Why do we want to expose those page objects in our API? Could we limit this lib to only provide actions?

dagguh: `internal` sounds good, we could expose them only when needed.
I tried it, but unfortunately, it has to be exposed for `IssueNavigatorPage` API:
image

If we hide/move `IssueNavigatorPage.waitForResults(results: IssueNavResultsView)`, then `IssueNavigatorPage` would lose its only non-deprecated API.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/issuenav/IssueNavResultsView.kt
dagguh: Hiding it would be nice. But it cannot be an `enum`, because:
* [enums block evolution](https://github.com/wyrzyk/java-api-design-checklist/blob/master/CHECKLIST.md#255-do-not-add-new-values-to-a-released-enumeration)
* [they get unwieldy very fast](https://github.com/atlassian/jira-actions/commit/6988b29451c9a840ccee8591ca481d6b513d5ec2#diff-d611ddd88e4934022a3e1e93fdd8d14e28636cab88bf1f8b04f76d2fa08f4a71L93-R104) (and this is before more methods like `listIssueKeys`
* there are more variations of interacting with the views than we think, e.g. we might detect them via the switcher icon rather than via the results

ewefie (approved):

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/issuenav/IssueNavResultsView.kt
pczuj: > enums block evolution

yup, I know, that's why I used "selection type similar to `enum`" wording 😉 

I wonder if the restriction around `enum` makes us unable to pass choices in non-polymorphic way or is there some other good way to do that. Not sure if passing `int` would be a good idea for use case like that - feels a bit cheap.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/issuenav/DetailView.kt
pczuj: Could we not deprecate the `waitForIssueNavigator()` method? It seems to be able to detect the view and wait for whatever is needed.

dagguh: The `UnknownView` impl irepresents a user, who has no list/detail view preferences and goes with whatever is currently selected. I thought it was a complication caused by old wrong assumptions and we don't need that anymore. Perhaps we still want to model such a user, but new API also allows that.

PR #119, opened 2023-03-27
Upload diagnoses

dagguh:
They contain browser recordings, which are useful to diagnose CI fails.


pczuj (approved):

dagguh:
Pushed a fix for `Artifact name is not valid: diagnoses/8.5.0. Contains the following character: Forward slash /`

ewefie (approved):

PR #120, opened 2023-03-27
JPERF-1050: Disable instance health notifications

dagguh:


src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/AdminAccess.kt
```diff
@@ -69,5 +69,16 @@ class AdminAccess(
         }
     }
 
+    internal fun runWithAccess(runnable: () -> Unit) {
```
pczuj, line 72: I don't have a good idea how to design a code that models web browser driver (still something I need to figure out), however I was thinking if this pattern shouldn't be used more often, e.g. when we open an menu (`administrate()`) maybe we should execute the operation (code) in the context of opened menu. 

In this sense maybe execution of this `runWithAccess` should be implicit inside `system()`, before it returns `SystemAdministrationPage`. Right now the page objects is returned, yet we don't land on the page.

pczuj (approved):
Looks good

src/main/kotlin/com/atlassian/performance/tools/jiraactions/administration/InstanceHealthNotificationsConfig.kt
```diff
@@ -0,0 +1,14 @@
+package com.atlassian.performance.tools.jiraactions.administration
+
+import org.openqa.selenium.By
+import org.openqa.selenium.WebElement
+import org.openqa.selenium.support.ui.Select
+
+internal class InstanceHealthNotificationsConfig(
+    private val config: WebElement
+) {
+    fun dontShowAny() {
```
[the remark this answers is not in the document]
dagguh: image
I choose to use user-facing language in our page objects and actions.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/AdminAccess.kt
dagguh: The wrapper was needed here, to express "before" (gain websudo) and "after" (drop websudo). Websudo needs to defer an action, but most UXes can just navigate away and forget about the past interactions.

pczuj: I understand that. 

Shouldn't this "before" and "after" wrapping also be applied to opening any of the administration pages? If I'm not wrong we shouldn't ever assume that there will be no websudo prompt when entering administration views. I'm also assuming that we never want to leave the websudo enabled when leaving the administration view, because otherwise we wouldn't do the websudo drop in the first place.

Maybe it's architecturally intentional for the page objects to be as atomic as possible? In such case I'd understand why we are not enabling sudo as part of view switch.

dagguh: Ah, using `runWithAccess` in each method of each admin page object? :thinking:

It wouldn't work, because when you're on `TroubleShootingAndSupportToolsPage` and then `AdminAccess.drop`, Jira redirects you back to the dashboard. So we'd have to nest it like this:
```
jira.runInAdministration {
    it.runInSystemAdministration {
        it.runInTroubleshootingPage {
            it.notifications().dontShowAny()
        }
    }
}
```
And count on the fact that only the top-level `runInAdministration` would actually `AdminAccess.grant` and `AdminAccess.drop` and the rest would fall out of `if (prompted)`. In runtime it's effectively the same as we have it currently.

pczuj: Maybe the nesting makes sense, but as mentioned I'm not yet opinionated about it.

For this change I thought more about something like this:
```
jira.administrate().system { systemAdministrationPage ->
    systemAdministrationPage
        .troubleshootingAndSupportTools()
        .instanceHealth()
        .notifications()
        .dontShowAny()
}
```
as a replacement for what is added to `SetUpAction`.

pczuj: Anyway this aspect doesn't affect API, so I will not block the change.

pczuj (approved):

PR #123, opened 2023-03-31
JPERF-1052 Change the expected condition in View History Tab action that is used to verify if the content is loaded.

ewefie:
Previously used 'stalenessOf' caused many timeout exceptions when clicking on the button was not successful. A new condition is more relaxed and when the button lacks the attribute indicating that the content is loading, the click will be repeated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,11 @@ Dropping a requirement of a major version of a dependency is a new contract.
 - Add `AdaptiveJqlMemory.Companion` JQL filtering extension functions.
   With them, you can actually replace the deprecated `SearchJql*Action`s.
 
+### Fixed
+- Change the condition used to verify if the content is loading in `ViewHistoryTabAction` to avoid TimeoutException. Fix [JPERF-1052].
+
+[JPERF-1052]: https://ecosystem.atlassian.net/browse/JPERF-1052
+
 ## [3.19.0] - 2023-03-30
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/HistoryTabPanel.kt
```diff
@@ -53,7 +53,7 @@ class HistoryTabPanel(
                 .perform()
             driver.wait(
                 duration,
-                ExpectedConditions.stalenessOf(loadMoreButton)
```
dagguh, line 56: Yeah, staleness is very tricky and not quite user-facing. Good removal

dagguh (approved):

CHANGELOG.md
```diff
@@ -26,6 +26,9 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/jira-actions/compare/release-3.19.0...master
 
+### Fixed
+- Change the condition used to verify if the content is loading in `ViewHistoryTabAction` to avoid TimeoutException. Fix [JPERF-1052].
+
```
dagguh, line 31:
```suggestion

[JPERF-1052]: https://ecosystem.atlassian.net/browse/JPERF-1052
```

dagguh (approved):

PR #126, opened 2023-04-03
Issue/jperf 1026 vus should timeout earlier

ewefie:
Change the default `WebDriver#wait` timeout to 10 seconds and use it consistently across all actions.


src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/page/IssueCreateDialog.kt
```diff
@@ -36,11 +34,7 @@ class IssueCreateDialog(
     }
 
     private fun waitForDialogToHide() {
-        driver.wait(
-            timeout = Duration.ofSeconds(30),
```
dagguh, line 40: Good

dagguh (approved):

pczuj:
Bold change. Let's see how it functions in our tests long term 🤞

dagguh:
It's time to put our foot down and say that 10s UX is unacceptable performance. No room for half-measures and excuses anymore.

PR #128, opened 2023-04-04
JPERF-1060: Close flags before using administration menu

pczuj:
Flags on top of the menu made the setup action to fails.

### DIFF
--- a/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/WebJira.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/WebJira.kt
@@ -107,6 +107,7 @@ data class WebJira(
     }
 
     internal fun administrate(): JiraAdministrationMenu {
+        NotificationPopUps(driver).waitUntilAuiFlagsAreGone()
         driver.findElement(By.id("admin_menu")).click()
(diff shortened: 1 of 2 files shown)


ewefie (approved):

dagguh (approved):

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/WebJira.kt
```diff
@@ -107,6 +107,7 @@ data class WebJira(
     }
 
     internal fun administrate(): JiraAdministrationMenu {
+        NotificationPopUps(driver).waitUntilAuiFlagsAreGone()
```
pczuj, line 110: Now I'm looking at this change again I think it can be problematic in the current form.

If I understand the code correctly it first waits up to 10 seconds for the flags to close themselves and only then it clicks them.

It means that when someone will want to use thi in an measured action it will have unpretictable results.

Currently it's only used in `SET_UP` measure which we don't judge upon, but it's fairly easy to use this menu in other measures without realising the problem. Should we safeguard future contributions in some way?

dagguh: Yeah, we should not wait for popups to appear, but close only the ones, which are already present.

PR #132, opened 2023-05-05
JPERF-1091 Move to the tab panel button before clicking it

ewefie:
Move to the tab panel button before clicking it to avoid intercepting clicks by the new sticky comment footer.

PR #133, opened 2023-05-08
JPERF-1099 Add Jira 9.8.0 to the matrix. 

ewefie:
Add Jira 9.8.0 to the matrix. Switch to gradle-build-actin where possible.


dagguh:
We expected this one to have red CI. But view issues weren't covered, so it went green. After #135 we've got the coverage, so let's rebase and rerun CI.

dagguh:
9.8.0 fails as expected. We can't merge a red test, so let's continue with the fix in #132.

PR #134, opened 2023-05-08
JPERF-1091: assert view issue

dagguh:


ewefie (approved):

dagguh:
Assertion fails as expected. The fix (with proper issue key) is here: https://github.com/atlassian/jira-actions/pull/135

PR #136, opened 2023-05-08
Issue/jperf 1098 test detail view too

dagguh:


src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/JiraCoreScenarioIT.kt
```diff
@@ -19,11 +21,20 @@ class JiraCoreScenarioIT : AbstractJiraCoreScenario() {
             .version(jiraVersion)
             .build()
             .provision().use { jira ->
-                val recordings = Paths.get("build/diagnoses/recordings/" + this::class.java.simpleName)
-
-                DockerisedChrome(recordings).start().use {browser ->
-                    shouldRunScenarioWithoutErrors(jira, browser.driver)
-                }
+                testScenario(jira, 123)
+                testScenario(jira, 456)
```
dagguh, line 25: After the seed fix, 123 picks detail view and 456 picks list view.

ewefie (approved):

src/test/kotlin/com/atlassian/performance/tools/jiraactions/api/scenario/JiraCoreScenarioIT.kt
dagguh: Very good question. I don't know right now.
I'll use made up numbers, but we can measure those when we get real about this.
If we have 2 hours of focus time a day, then a 20 minute dev loop gives us max 6 experiments a day. A 30 minute dev loop would cap it to 4. We know we'd like to be more productive, so the current thruput is not enough, we need more, say 20. How to increase it? Many ways:
* adding concurrency across `testScenario(jira, seed)` could save 1 minute on my machine, 2 minutes for @ewefie, 3 minutes on CI
* caching `com.atlassian.performance.tools.dockerinfrastructure.jira.SetUpFromScratchAction` could save 12 minutes, but it could cost more dev-time/complexity than concurrency

PR #140, opened 2023-05-12
JPERF-1107: Introduce a test about obscured issue create button

pczuj:
This test reproduces the problem that we've noticed in our regular tests. See: https://ecosystem.atlassian.net/browse/JPERF-1107

Next steps: 
 * Fix the test
 * Refactor the test


pczuj:
I was running this test 2 times and it took 13 and 6 min. This is a significant overhead and I believe most of it comes from Jira setup time. Maybe if there would be more similar action tests we could reuse Jira between them?

PR #154, opened 2023-10-04
Issue/jperf 1395 fix api dependencies

pczuj:


pczuj:
Without it I'd need to explicitly depend on the `json-api` to define own `ActionType`. Otherwise the compiler complains as I can't really use e.g. `SearchJqlObservation`.

dagguh:
Thanks for a nice commit message

dagguh (approved):

PR #156, opened 2023-10-24
JPERF-1408: Record PerformanceServerTiming

dagguh:


src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/w3c/PerformanceResourceTiming.kt
```diff
@@ -23,5 +23,12 @@ class PerformanceResourceTiming internal constructor(
...
     val encodedBodySize: Long,
-    val decodedBodySize: Long
+    val decodedBodySize: Long,
+    /**
+     * Represents the [serverTiming attribute](https://www.w3.org/TR/2023/WD-server-timing-20230411/#servertiming-attribute)
+     *
+     * @return Null if blocked, e.g. by same-origin policy.
+     *         Empty if not blocked, but server didn't send any `Server-Timing` headers.
+     */
+    val serverTiming: List<PerformanceServerTiming>?
```
mgrzaslewicz, line 33: maybe better wrap it in `ServerTiming` class and provide a reason there plus non nullable list
I can see more unknown reasons incoming

mgrzaslewicz (changes requested):
consider non-nullable type

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/w3c/PerformanceResourceTiming.kt
mgrzaslewicz: After discussion, nullable field is fine and enough

mgrzaslewicz (approved):

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/w3c/PerformanceResourceTiming.kt
dagguh: That's how the server timing standard does it, I'm just echoing the web standard.

PR #159, opened 2023-10-25
JPERF-1401: Reuse `JsonProvider`

dagguh:
Avoid scanning the classpath on every JSON value creation.

Repeat the same perf fix as in https://github.com/atlassian/report/pull/97 I apologize for not adding a test for the APIs in this lib. Our team is running out of time.
Dear future reader, if the libs get development time again, I would:
* import a big action metrics file with lots of drilldown JSONs
* run parsing and serialization in parallel with concurrency = 4


dagguh:
- [ ] CHANGELOG

mgrzaslewicz (approved):

PR #160, opened 2023-11-07
Add nodeId to performance entries

mgrzaslewicz:


mgrzaslewicz:
Closing. Removing deprecations is too much as it impacts other libraries. Should be addressed in a commit that focuses only on deprecations

PR #161, opened 2023-11-08
Add nodeId and requestId to performance entries

mgrzaslewicz:


src/main/kotlin/com/atlassian/performance/tools/jiraactions/w3c/harvesters/JsResources.kt
```diff
@@ -63,12 +66,31 @@ private fun parsePerformanceEntry(
...
+    val result = jsServerTimings.map { parseServerTiming(it) }.toMutableList()
+    val responseHeaders = getResponseHeaders(jsExecutor)
+    val nodeId = responseHeaders["x-anodeid"]
+    val requestId = responseHeaders["x-arequestid"]
+    addAttribute(result, "nodeId", nodeId)
+    addAttribute(result, "requestId", requestId)
+    return result
+}
+
+fun addAttribute(result: MutableList<PerformanceServerTiming>, attributeName: String, attributeValue: String?) {
```
pczuj, line 84: Should it be public?

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/w3c/RecordedPerformanceEntries.kt
```diff
@@ -4,7 +4,7 @@ package com.atlassian.performance.tools.jiraactions.api.w3c
  * Holds interesting recorded performance entries.
  * They share the same timeline and can be cross-examined.
  */
-class RecordedPerformanceEntries internal constructor(
+data class RecordedPerformanceEntries internal constructor(
```
pczuj, line 7: Could you explain why we want/need the data class here?

src/main/kotlin/com/atlassian/performance/tools/jiraactions/w3c/harvesters/JsResources.kt
```diff
@@ -63,12 +66,31 @@ private fun parsePerformanceEntry(
...
     if (jsServerTimings !is List<*>) {
         throw Exception("Unexpected non-list JavaScript value: $jsServerTimings")
     }
-    return jsServerTimings.map { parseServerTiming(it) }
+    val result = jsServerTimings.map { parseServerTiming(it) }.toMutableList()
+    val responseHeaders = getResponseHeaders(jsExecutor)
+    val nodeId = responseHeaders["x-anodeid"]
+    val requestId = responseHeaders["x-arequestid"]
+    addAttribute(result, "nodeId", nodeId)
+    addAttribute(result, "requestId", requestId)
```
pczuj, line 80: Not sure if I like the fact that we are artificially putting product specific data into W3C model and that it happens in `w3c` package, not in some higher level wrapper. This creates a bit of a chaos of where things belong.

The main problem with this that I see is that if the Jira specific data will become API we won't be able to change our mind about this model. The only way will be fork or API break.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/BackendTimeSlot.kt
```diff
@@ -0,0 +1,16 @@
+package com.atlassian.performance.tools.jiraactions.api
+
+import java.time.Duration
+import java.time.Instant
+
+data class BackendTimeSlot internal constructor(
```
pczuj, line 6: Personally I'd prefer the data class methods to be explicit when they are becoming API. Otherwise it's easy to unintentionally miss this when changing the class.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/api/w3c/RecordedPerformanceEntries.kt
mgrzaslewicz: There were need for hashCode/equals and toString in tests hence `data class` doing what we need.
Plus copy() which we don't need. I might just implement those 3 methods as well

CHANGELOG.md
```diff
@@ -26,6 +26,10 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/jira-actions/compare/release-3.24.0...master
 
+### Added
+- Add `nodeId` to `PerformanceServerTiming`. Aid with [JPERF-1409].
```
pczuj, line 30: Could we make this a lib's internal contract and not expose it in `PerformanceServerTiming`? We'd expose it only in `BackendTimeSlot`.

src/main/kotlin/com/atlassian/performance/tools/jiraactions/w3c/harvesters/JsResources.kt
mgrzaslewicz: our API is `BackendTimeSlot`, javascript field is a technical detail

CHANGELOG.md
mgrzaslewicz: I'll remove the javascript detail from changelog

pczuj (approved):

